PR #33, opened 2023-09-01
Handle vehicle burn and write tests

0xdev22:


migrations/00012_vehicles_privileges_cascade_delete.sql
```diff
@@ -1,7 +1,7 @@
 -- +goose Up
 -- +goose StatementBegin
 ALTER TABLE privileges
-    DROP CONSTRAINT IF EXISTS privileges_vehicle_token_id_fkey,
+    DROP CONSTRAINT vehicle_privileges_vehicle_token_id,
```
elffjs, line 4: Can we rename the token_id column to vehicle_token_id maybe? Doesn’t have to be in this PR.

elffjs (approved):

PR #34, opened 2023-09-07
Listen for and handle DCN events, persist DCN events

0xdev22:
image


gqlgen.yml
```diff
@@ -87,6 +87,8 @@ models:
       - github.com/99designs/gqlgen/graphql.Int32
   Address:
     model: github.com/DIMO-Network/identity-api/graph/types.Address
+  Byte:
```
elffjs, line 90: Can we call it `Bytes`? I think this is more accurate.

graph/schema.graphqls
```diff
@@ -152,3 +154,9 @@ type SyntheticDevice {
   address: Address!
   mintedAt: Time!
 }
+
+type DCNConnection {
```
elffjs, line 158: Why is it `DCNConnection`? That’s a bit strange. Maybe just `Dcn`?

gqlgen.yml
elffjs: Actually it’s called `Bytes` everywhere else. How is this working?

internal/config/settings.go
```diff
@@ -12,4 +12,5 @@ type Settings struct {
 	DIMORegistryAddr      string      `yaml:"DIMO_REGISTRY_ADDR"`
 	VehicleNFTAddr        string      `yaml:"DIMO_VEHICLE_NFT_ADDR"`
 	AftermarketDeviceAddr string      `yaml:"AFTERMARKET_DEVICE_CONTRACT_ADDRESS"`
+	DimoDCNRegistryAddr   string      `yaml:"DIMO_DCN_REGISTRY_ADDR"`
```
elffjs, line 15: The “D” in DCN stands for DIMO, so I think we can leave out `DIMO_`.

internal/services/event_data_models.go
```diff
@@ -90,3 +90,18 @@ type DeviceDefinition struct {
 		Year  int
 	}
 }
+
+type NewDCNNodeEventData struct {
+	Node  []byte
+	Owner common.Address
+}
+
+type NewDCNResolverEventData struct {
```
elffjs, line 99: Let’s actually not handle this. We’ll just look at the default resolver and ignore everything else.

internal/services/contracts_dcn_handler.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (c *ContractsEventsConsumer) handleNewDCNExpiration(ctx context.Context, e *ContractEventData) error {
+	logger := c.log.With().Str("EventName", NewExpiration.String()).Logger()
+
+	var args NewDCNExpirationEventData
+	if err := json.Unmarshal(e.Arguments, &args); err != nil {
+		return err
+	}
+
+	dcn, err := models.DCNS(
```
elffjs, line 69: I don’t think you need to actually query for this. We can construct the row with the node (the primary key) and the expiration time and call `Update`, is my guess.

elffjs (changes requested):
Minor things.

internal/services/contracts_dcn_handler.go
```diff
@@ -0,0 +1,87 @@
+package services
+
```
elffjs, line 2: I like that you split these out.

graph/types/blockchain.go
```diff
@@ -26,6 +26,21 @@ func UnmarshalAddress(v interface{}) (common.Address, error) {
 	}
 }
 
+func MarshalBytes(b []byte) graphql.Marshaler {
+	return graphql.WriterFunc(func(w io.Writer) {
+		_, _ = io.WriteString(w, strconv.Quote(string(b)))
+	})
+}
+
+func UnMarshalBytes(b interface{}) ([]byte, error) {
```
elffjs, line 35: This looks very different from the other type. I would be surprised if this worked.

elffjs, line 35: Is this the correct spelling?

elffjs (changes requested):

internal/services/contracts_dcn_handler_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	dcn, err := models.DCNS().All(o.ctx, o.pdb.DBS().Reader.DB)
+	o.NoError(err)
+
+	o.Len(dcn, 1)
+	o.Equal(eventData.Node, dcn[0].Node)
+	o.Equal(owner.Bytes(), dcn[0].OwnerAddress)
+	o.Equal(addr.Bytes(), dcn[0].ResolverAddress.Bytes)
+	o.Equal(currTime, dcn[0].Expiration.Time)
+}
```
elffjs, line 203: I think one gap in these tests is that we’re not checking the serialization, and I suspect there’s something wrong there.

graph/types/blockchain.go
```diff
@@ -26,6 +26,21 @@ func UnmarshalAddress(v interface{}) (common.Address, error) {
...
+	byt, ok := b.(string)
+	if !ok {
+		return []byte{}, fmt.Errorf("value must be a string")
```
elffjs, line 38: Can return `nil` instead of an empty array.

elffjs:
Could we put a screenshot in? That would raise my confidence?

I don’t have the files for this in the event simulator, but we could put them in there and I could show you how

graph/schema.graphqls
```diff
@@ -152,3 +154,9 @@ type SyntheticDevice {
...
+  node: Bytes!
+  owner: Address!
+  expiresAt: Time!
```
elffjs, line 161: Could we add `mintedAt` as well?

graph/schema.resolvers.go
```diff
@@ -35,6 +36,11 @@ func (r *queryResolver) Vehicle(ctx context.Context, id int) (*model.Vehicle, er
 	return r.Repo.GetVehicle(ctx, id)
 }
 
+// Dcn is the resolver for the dcn field.
+func (r *queryResolver) Dcn(ctx context.Context, node string) (*model.DCNConnection, error) {
+	panic(fmt.Errorf("not implemented: Dcn - dcn"))
```
elffjs, line 41: 🚨

graph/types/blockchain.go
```diff
@@ -26,6 +27,25 @@ func UnmarshalAddress(v interface{}) (common.Address, error) {
 	}
 }
 
+func MarshalBytes(b []byte) graphql.Marshaler {
+	return graphql.WriterFunc(func(w io.Writer) {
+		out, _ := json.Marshal(b)
```
elffjs, line 32: We can use `hexutil.Encode` to avoid incurring the JSON overhead. Also, when we're doing JSON stuff I'd like to keep using the one, faster package we're using elsewhere.

graph/types/blockchain.go
```diff
@@ -26,6 +27,25 @@ func UnmarshalAddress(v interface{}) (common.Address, error) {
...
+		_, _ = io.WriteString(w, string(out))
+	})
+}
+
+func UnmarshalBytes(v interface{}) ([]byte, error) {
+	switch v := v.(type) {
+	case string:
+		out, _ := json.Marshal(v)
```
elffjs, line 40: This seems unlikely to work. `v` here is coming from naive JSON parsing: it ought to be the hex-encoded string, so something like `0xaf4b`. And then I think we're going to get the ASCII bytes for `"0xaf4b"`.

internal/repositories/dcn.go
```diff
@@ -0,0 +1,36 @@
...
+
+func DCNToAPI(d *models.DCN) *gmodel.Dcn {
+	return &gmodel.Dcn{
+		Owner: common.BytesToAddress(d.OwnerAddress),
+		Node:  d.Node,
+	}
+}
+
+func (r *Repository) GetDCNByNode(ctx context.Context, node []byte) (*gmodel.Dcn, error) {
+	var n []byte
```
elffjs, line 21: Why should we have to do this again? At this point, `node` should be the bytes we want.

internal/repositories/dcn.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	dcn, err := models.DCNS(
+		models.DCNWhere.Node.EQ(n),
+	).One(ctx, r.pdb.DBS().Reader)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Println(dcn)
```
elffjs, line 33: Weeu.

internal/services/contracts_dcn_handler.go
elffjs: I made this change. Things seem to pass.

migrations/00013_dcn_table.sql
```diff
@@ -0,0 +1,14 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE dcn (
```
elffjs, line 3: I think this should probably be `dcns`. I'll do it.

elffjs:
I made a bunch of changes. In general I think blockchain people want to see hexadecimal, so we are going to give it to them. I added a test that exercises the encoding stuff too, take a look.

elffjs:
I don't think the test is perfect, but would you like to add the expiration event/response to it and we call this PR good?

elffjs:
Actually, I want to add the mint time. I'll do that.

elffjs (approved):

PR #35, opened 2023-09-12
Handle NameChangedEvent and generate new models to support it

0xdev22:


internal/repositories/dcn.go
```diff
@@ -13,6 +13,7 @@ func DCNToAPI(d *models.DCN) *gmodel.Dcn {
 		Owner:     common.BytesToAddress(d.OwnerAddress),
 		Node:      d.Node,
 		ExpiresAt: d.Expiration.Ptr(),
+		Name:      &d.Name.String,
```
elffjs, line 16: Again, `d.Name.Ptr()` is probably what you want.

graph/dcn_test.go
```diff
@@ -113,4 +114,35 @@ func TestDCNQuery(t *testing.T) {
 
 	assert.NoError(err)
 	assert.Equal(expected, currTime)
+
```
elffjs, line 117: We probably want to not have a big mega-test here, but this is okay for now.

graph/dcn_test.go
```diff
@@ -113,4 +114,35 @@ func TestDCNQuery(t *testing.T) {
...
+				node
+				owner
+				expiresAt
+				name
+			}
+		}
+	`, &dcnr, client.Var("node", "0x6665255d9e2cfc4ed6d3064285c4d20c864af81dc0d26c94c5380f67f9dd57ed"))
+
+	assert.NoError(err)
+	assert.Equal(*dcnr.DCN.Name, mockName)
```
elffjs, line 147: I think these are `Assert(expected, actual)`.

elffjs (changes requested):

internal/services/contracts_dcn_handler.go
```diff
@@ -55,3 +55,27 @@ func (c *ContractsEventsConsumer) handleNewDCNExpiration(ctx context.Context, e
...
+		Node: args.Node,
+		Name: null.StringFrom(args.Name),
+	}
+
+	_, err := dcn.Update(ctx, c.dbs.DBS().Writer, boil.Whitelist(models.DCNColumns.Name))
+	if err != nil {
+		return err
+	}
+
+	logger.Info().Str("Node", string(args.Node)).Msg(eventName + " Event processed successfuly")
```
elffjs, line 78: `string(node)` is usually not going to be meaningful. Do you want to try `.Bytes("node", args.Node)`?

elffjs: Alternatively just something like [`hexutil.Encode`](https://pkg.go.dev/github.com/ethereum/go-ethereum/common/hexutil#Encode).

graph/dcn_test.go
```diff
@@ -113,4 +114,35 @@ func TestDCNQuery(t *testing.T) {
...
+			dcn(node: $node) {
+				node
+				owner
+				expiresAt
+				name
+			}
+		}
+	`, &dcnr, client.Var("node", "0x6665255d9e2cfc4ed6d3064285c4d20c864af81dc0d26c94c5380f67f9dd57ed"))
+
+	assert.NoError(err)
```
elffjs, line 146: Does this do anything?

graph/dcn_test.go
```diff
@@ -74,12 +75,14 @@ func TestDCNQuery(t *testing.T) {
...
 				owner
 				expiresAt
+				name
 			}
 		}
 	`, &dcnr, client.Var("node", "0x6665255d9e2cfc4ed6d3064285c4d20c864af81dc0d26c94c5380f67f9dd57ed"))
 
 	assert.Equal("0x6665255d9e2cfc4ed6d3064285c4d20c864af81dc0d26c94c5380f67f9dd57ed", dcnr.DCN.Node)
 	assert.Equal("0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045", dcnr.DCN.Owner)
+	assert.Nil(dcnr.DCN.Name)
```
elffjs, line 85: Added this to have a failing test. Let's make this pass!

internal/services/contracts_dcn_handler.go
elffjs: Like it produces stuff like

```
{"level":"info","EventName":"NewNode","Node":"fe%]\ufffd,\ufffdN\ufffd\ufffd\u0006B\ufffd\ufffd\ufffd\f\ufffdJ\ufffd\u001d\ufffd\ufffdl\ufffd\ufffd8\u000fg\ufffd\ufffdW\ufffd","message":"NewNode Event processed successfuly"}
```

internal/services/contracts_events_consumer.go
```diff
@@ -123,6 +124,8 @@ func (c *ContractsEventsConsumer) Process(ctx context.Context, event *shared.Clo
 			return c.handleNewDcnNode(ctx, &data)
 		case NewExpiration:
 			return c.handleNewDCNExpiration(ctx, &data)
+		case NameChanged:
```
elffjs, line 127: Bigger problem: this event does not come from the DCN registry, it comes from the resolver. The registry does not know about names.

elffjs: So in the test, using the addresses from prod, the NameChanged event would come from [0x60627326F55054Ea448e0a7BC750785bD65EF757](https://polygonscan.com/address/0x60627326F55054Ea448e0a7BC750785bD65EF757#events)

https://polygonscan.com/address/0x60627326F55054Ea448e0a7BC750785bD65EF757#events

elffjs: I will try to get this running locally so you can see for yourself. No promises

elffjs: Registry events, for comparison

https://polygonscan.com/address/0xE9F4dfE02f895DC17E2e146e578873c9095bA293#events

elffjs: In general I think when writing these it’s good to use the “test data” that we have, namely the live deployments 

https://github.com/DIMO-Network/web3-deploys

elffjs (approved):

PR #38, opened 2023-09-18
Si 2042 dcn by name

0xdev22:
image

image

image

image


elffjs:
Is this a draft?

I like the idea of `@oneOf` as an advisory to the client. I think implementing it in a generic way is going to be difficult. Your middleware here is super specific to the situation at hand.

I would sort this out in the resolver for now.

graph/schema.graphqls
```diff
@@ -36,7 +43,7 @@ type Query {
   ): AftermarketDeviceConnection!
   vehicle(id: Int!): Vehicle!
 
-  dcn(node: Bytes!): DCN!
+  dcn(input: DCNInput!): DCN!
```
elffjs, line 46: I like the naming from [the RFC](https://github.com/graphql/graphql-spec/pull/825). `dcn(by: DCNBy!)`.

graph/schema.graphqls
```diff
@@ -12,8 +12,19 @@ scalar Address
 An integer of arbitrary precision, decimal-encoded. Typically a uint256.
 """
 scalar BigInt
+
+"""
+A general byte representation
```
elffjs, line 17: It looks like we need to merge main in.

graph/schema.graphqls
```diff
@@ -12,8 +12,19 @@ scalar Address
...
+"""
 scalar Bytes
 
+directive @oneOf on OBJECT | INPUT_OBJECT | FIELD_DEFINITION
```
elffjs, line 21: Do we need anything besides `INPUT_OBJECT`?

cmd/identity-api/main.go
```diff
@@ -51,9 +52,14 @@ func main() {
 
 	repo := repositories.New(dbs)
 
-	s := handler.NewDefaultServer(graph.NewExecutableSchema(graph.Config{Resolvers: &graph.Resolver{
+	cfg := graph.Config{Resolvers: &graph.Resolver{
 		Repo: repo,
-	}}))
+	}}
+	cfg.Directives.OneOf = func(ctx context.Context, obj interface{}, next graphql.Resolver) (res interface{}, err error) {
```
elffjs, line 58:
```suggestion
	cfg.Directives.OneOf = func(ctx context.Context, _ any, _ graphql.Resolver) (any, error) {
```
Should we also leave a comment like "The directive on its own is advisory; everything is enforced inside of the resolver".

internal/repositories/dcn.go
```diff
@@ -14,14 +17,44 @@ func DCNToAPI(d *models.DCN) *gmodel.Dcn {
 		Node:      d.Node,
 		ExpiresAt: d.Expiration.Ptr(),
 		Name:      d.Name.Ptr(),
+		VehicleID: d.VehicleID.Ptr(),
 	}
 }
 
-func (r *Repository) GetDCNByNode(ctx context.Context, node []byte) (*gmodel.Dcn, error) {
-	dcn, err := models.DCNS(models.DCNWhere.Node.EQ(node)).One(ctx, r.pdb.DBS().Reader)
+func (r *Repository) getDCNByNode(ctx context.Context, dcnParams model.DCNBy) (*gmodel.Dcn, error) {
```
elffjs, line 24: I think it's a bit strange that these accept `model.DCNBy`. That seems to invite trouble.

Could we have this one accept `[]byte` and the next one accept `string`?

internal/repositories/dcn.go
```diff
@@ -14,14 +17,44 @@ func DCNToAPI(d *models.DCN) *gmodel.Dcn {
...
+		models.DCNWhere.Name.EQ(null.StringFrom(*dcnParams.Name)),
+	).One(ctx, r.pdb.DBS().Reader)
+	if err != nil {
+		return nil, err
+	}
+
+	return DCNToAPI(dcn), nil
+}
+
+func (r *Repository) GetDCN(ctx context.Context, params model.DCNBy) (*gmodel.Dcn, error) {
```
elffjs, line 46: The repository seems like a strange place for this. Could we do it in the resolver?

internal/repositories/dcn.go
```diff
@@ -14,14 +17,44 @@ func DCNToAPI(d *models.DCN) *gmodel.Dcn {
...
+	if params.Name != nil && len(params.Node) > 0 {
+		return nil, fmt.Errorf("please provide one of Name or Node but not both")
```
elffjs, line 48: This is terribly picky but

1. I wouldn't include "please" in a repository error
2. We don't need `fmt` because we're not interpolating anything

graph/schema.resolvers.go
```diff
@@ -36,8 +44,8 @@ func (r *queryResolver) Vehicle(ctx context.Context, id int) (*model.Vehicle, er
 }
 
 // Dcn is the resolver for the dcn field.
-func (r *queryResolver) Dcn(ctx context.Context, node []byte) (*model.Dcn, error) {
-	return r.Repo.GetDCNByNode(ctx, node)
+func (r *queryResolver) Dcn(ctx context.Context, by model.DCNBy) (*model.Dcn, error) {
+	return r.Repo.GetDCN(ctx, by)
```
elffjs, line 48: Can we put more logic here? I'd also like to check that `node` has length 32.

elffjs (changes requested):
I want to take another look after main is merged in. It's hard to navigate this right now.

I appreciate the examples.

internal/repositories/dcn.go
0xdev22: I view the resolver more like a Router in the MVC architecture. Moving this logic there would mean its involved in validation, which is a logic i don't think belongs in a router, since its basically meant to just make sure the right handler gets called.

graph/schema.resolvers.go
0xdev22: https://github.com/DIMO-Network/identity-api/pull/38#discussion_r1331956449

elffjs: Are you checking the length?

internal/repositories/dcn.go
elffjs: You're putting it in the repository, the storage layer. If you want to introduce another layer, that's fine.

In the interest of getting this moving—if the two functions this calls had narrower input types, I'd be happier.

graph/schema.resolvers.go
0xdev22: It was more about not checking the length in the resolver.

internal/repositories/dcn.go
elffjs: So, it has to go one of two bad places. I have a preference, you have a preference, that's fine; what I don't like is `GetByName(BigUnionType)`.

graph/schema.resolvers.go
elffjs: Are we checking it anywhere?

0xdev22: Yeah, will check in the resolver

elffjs (approved):

PR #43, opened 2023-09-28
Implement filtering for privileges

0xdev22:


graph/schema.graphqls
```diff
@@ -159,12 +167,7 @@ type Vehicle implements Node {
   """
   A Relay-style connection listing any active privilege grants on this vehicle.
   """
-  privileges(
-    first: Int
-    after: String
-    last: Int
-    before: String
```
elffjs, line 166: We have to preserve these in order to follow Relay. The filter should be in addition to these.

elffjs (changes requested):
This has filtering by user but not filtering by privilege, and it has no tests.

elffjs:
> This has filtering by user but not filtering by privilege, and it has no tests.

Are we ignoring this?

elffjs:
I don't think this does what was asked, but I don't want us to spend more time on it so I'm just going to merge it and cut down the scope of the issue.

PR #44, opened 2023-10-02
Update schema.graphqls

zer0stars:
Add descriptions from chatgpt.


elffjs:
I would rather sit down and write. I will do this today.

PR #47, opened 2023-10-12
Add short name to AfterMarket and Vehicles

0xdev22:


elffjs (changes requested):

elffjs:
I promise to write a version of this three words thing that can't error. It shouldn't error.

internal/helpers/helpers.go
```diff
@@ -31,3 +35,20 @@ func GenerateDCNNode() []byte {
...
+	if err != nil {
+		return "", err
+	}
+	name := strings.Join(mn, " ")
+
+	return name, nil
+}
+
+func IntToBytes(intVal int) []byte {
+	b := make([]byte, 32)
```
elffjs, line 50: It doesn't really matter, but note that the `32` in `Uint32` is bits, not bytes. Anyway, none of this matters because only the first 5 bytes are relevant to us.

internal/helpers/helpers.go
```diff
@@ -31,3 +35,20 @@ func GenerateDCNNode() []byte {
 
 	return b
 }
+
+func CreateMnemonic(addr []byte) (string, error) {
```
elffjs, line 39: I think the name `addr` is misleading in the case of vehicles.

elffjs (approved):
Just going to merge this and fix my qualms later.

PR #49, opened 2023-10-17
Process rewards event for connectionStreak, aftermarketDevice and syntheticDevice

0xdev22:


migrations/00021_rewards.sql
```diff
@@ -0,0 +1,22 @@
+-- +goose Up
```
elffjs, line 1: We should think about these names for a bit.

internal/services/contracts_events_consumer.go
```diff
@@ -521,3 +527,61 @@ func (c *ContractsEventsConsumer) handleSyntheticDeviceNodeBurnedEvent(ctx conte
...
+	return nil
+}
+
+func (c *ContractsEventsConsumer) handleTokensTransferredForConnectionStreak(ctx context.Context, e *ContractEventData) error {
+	var args TokensTransferredForConnectionStreakData
+	if err := json.Unmarshal(e.Arguments, &args); err != nil {
+		return err
+	}
+
+	reward, err := models.Rewards(
```
elffjs, line 571: Must we query? I think we could construct a struct with issuance week, vehicle id, streak earning, connection streak, and call update on that.

internal/services/contracts_events_consumer.go
```diff
@@ -521,3 +527,61 @@ func (c *ContractsEventsConsumer) handleSyntheticDeviceNodeBurnedEvent(ctx conte
...
 }
+
+func (c *ContractsEventsConsumer) handleTokensTransferredForDevice(ctx context.Context, e *ContractEventData) error {
+	var args TokensTransferredForDeviceData
+	if err := json.Unmarshal(e.Arguments, &args); err != nil {
+		return err
+	}
+
+	reward := models.Reward{
+		IssuanceWeek:      1,
```
elffjs, line 538: Just assume the event has this. We should either do that issue first or invent a name now. We can always change it later.

internal/services/contracts_events_consumer.go
```diff
@@ -521,3 +527,61 @@ func (c *ContractsEventsConsumer) handleSyntheticDeviceNodeBurnedEvent(ctx conte
...
+		VehicleID:         int(args.VehicleID.Int64()),
+		ReceivedByAddress: null.BytesFrom(args.User.Bytes()),
+		EarnedAt:          e.Block.Time,
+	}
+
+	cols := models.RewardColumns
+
+	if common.HexToAddress(c.settings.AftermarketDeviceAddr) == args.DeviceNftProxy {
```
elffjs, line 546: Upserts look good.

sqlboiler.toml
```diff
@@ -2,7 +2,7 @@
 dbname = "identity_api"
 blacklist = ["migrations"]
 host   = "localhost"
-port   = 5432
+port   = 5433
```
elffjs, line 5: Remember to not commit this.

settings.sample.yaml
```diff
@@ -13,3 +13,4 @@ DIMO_REGISTRY_CHAIN_ID: 31337
 DIMO_REGISTRY_ADDR: "0x5FbDB2315678afecb367f032d93F642f64180aa3"
 DIMO_VEHICLE_NFT_ADDR: "0xc6e7DF5E7b4f2A278906862b61205850344D4e7d"
 AFTERMARKET_DEVICE_CONTRACT_ADDRESS: "0xcf9af64522162da85164a714c23a7705e6e466b3"
+SYNTHETIC_DEVICE_CONTRACT_ADDRESS: 0x85226A67FF1b3Ec6cb033162f7df5038a6C3bAB2
```
elffjs, line 16: I would quote this to be safe. YAML does weird things to hex.

Makefile
```diff
@@ -70,3 +72,14 @@ clean:
...
+run: ## Run the app.
+	go run ./cmd/identity-api
+migrate: ## Run database migrations.
+	go run ./cmd/identity-api migrate
+sql: ## Create a new SQL migration file. Use the NAME variable to set the name: "make sql NAME=dcn_table".
+	goose -dir migrations create $(NAME) sql
+boil: ## Generate SQLBoiler models.
+	sqlboiler psql --no-tests --wipe
+gql: ## Generate gqlgen code.
+	go run github.com/99designs/gqlgen generate
```
elffjs, line 85: Newline.

migrations/00021_rewards.sql
```diff
@@ -0,0 +1,22 @@
...
+-- +goose StatementBegin
+CREATE TABLE rewards (
+    issuance_week int NOT NULL,
+    connection_streak int,
+    streak_earning int,
```
elffjs, line 6: `streak_earning` vs `aftermarket_earnings`. We should be consistent.

migrations/00021_rewards.sql
```diff
@@ -0,0 +1,22 @@
...
+    vehicle_id int NOT NULL CONSTRAINT rewards_vehicle_token_id REFERENCES vehicles (id),
+    aftermarket_token_id int CONSTRAINT rewards_aftermarket_token_id REFERENCES aftermarket_devices (id),
+    aftermarket_earnings int,
+    synthetic_token_id int CONSTRAINT rewards_synthetic_token_id REFERENCES synthetic_devices (id),
+    synthetic_earnings int,
+    received_by_address bytea CONSTRAINT received_by_address_check CHECK (length(received_by_address) = 20),
+    earned_at timestamptz(0) NOT NULL,
+
```
elffjs, line 14: I'm just going to do the renames I want.

internal/services/contract_events_consumer_test.go
```diff
@@ -1009,3 +1010,540 @@ func Test_HandleVehicle_Transferred_To_Zero_Event_NoDelete_AfterMarketDevice(t *
...
+
+	err = contractEventConsumer.Process(ctx, &e)
+	assert.NoError(t, err)
+
+	reward, err := models.Rewards().All(ctx, pdb.DBS().Reader)
+	assert.NoError(t, err)
+
+	assert.Len(t, reward, 1)
+
+	if len(reward) > 0 {
```
elffjs, line 1195: This is neat.

internal/services/event_data_models.go
```diff
@@ -127,3 +127,18 @@ type DCNVehicleIdChangedData struct {
...
+	Amount         *big.Int       `json:"_amount"`
+	VehicleID      *big.Int       `json:"vehicleNodeId"`
+	DeviceNftProxy common.Address `json:"deviceNftProxy"`
+	DeviceNode     *big.Int
+}
+
+type TokensTransferredForConnectionStreakData struct {
+	User             common.Address
+	Amount           *big.Int `json:"_amount"`
+	VehicleID        *big.Int `json:"vehicleNodeId"`
```
elffjs, line 142: Can we just keep the names the same? I don't see the need for this.

internal/services/contract_events_consumer_test.go
```diff
@@ -1009,3 +1010,540 @@ func Test_HandleVehicle_Transferred_To_Zero_Event_NoDelete_AfterMarketDevice(t *
...
+
+	err = afterMarketDevice.Insert(ctx, pdb.DBS().Writer, boil.Infer())
+	assert.NoError(t, err)
+
+	err = syntheticDevice.Insert(ctx, pdb.DBS().Writer, boil.Infer())
+	assert.NoError(t, err)
+
+	contractEventConsumer := NewContractsEventsConsumer(pdb, &logger, &settings)
+
+	payload := []struct {
```
elffjs, line 1466: `payloads` plural?

internal/services/contracts_events_consumer.go
elffjs: Or you could use https://github.com/DIMO-Network/rewards-api/blob/046c4a852e89ce4d8dc1bf62a544a40cea1a45e8/internal/services/rewards.go#L95

But I would really like to put it in the event.

migrations/00021_rewards.sql
```diff
@@ -0,0 +1,22 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE rewards (
+    issuance_week int NOT NULL,
+    connection_streak int,
+    streak_earning int,
+    vehicle_id int NOT NULL CONSTRAINT rewards_vehicle_token_id REFERENCES vehicles (id),
+    aftermarket_token_id int CONSTRAINT rewards_aftermarket_token_id REFERENCES aftermarket_devices (id),
+    aftermarket_earnings int,
```
elffjs, line 9: Unfortunately I don’t think `int` (IIRC this is 32 bits) is going to cut it. The numbers are things like 14 x 10^18. We’re probably back to numeric.

I think it is okay to have these token fields be non-null.

elffjs (changes requested):

Makefile
0xdev22: Not sure I understand

elffjs: Newlines at the end of files! I'll just do it.

https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

migrations/00021_rewards.sql
elffjs: This is still a problem. For example, look at the transfers from the start of this week:

https://polygonscan.com/tx/0x2298e26e8c50fcf7dcebb656e99f05f5af75d992ae7345602aafd968806d3609#eventlog

14054590318231538246! Very large.

internal/services/contracts_events_consumer.go
```diff
@@ -126,6 +128,10 @@ func (c *ContractsEventsConsumer) Process(ctx context.Context, event *shared.Clo
 			return c.handleVehicleTransferEvent(ctx, &data)
 		case PrivilegeSet:
 			return c.handlePrivilegeSetEvent(ctx, &data)
+		case TokensTransferredForDevice:
+			return c.handleTokensTransferredForDevice(ctx, &data)
+		case TokensTransferredForConnectionStreak:
+			return c.handleTokensTransferredForConnectionStreak(ctx, &data)
```
elffjs, line 134: Actually, this is suspicious—shouldn't these be coming from the rewards contract? I think we need to add a setting.

migrations/00021_rewards.sql
0xdev22: This can't be non-null since it will error when any of the tokeniD's isn't present and I try to create any records

elffjs: I mean they can be zero. Nil is fine for now.

settings.sample.yaml
```diff
@@ -14,3 +14,4 @@ DIMO_REGISTRY_ADDR: "0x5FbDB2315678afecb367f032d93F642f64180aa3"
 DIMO_VEHICLE_NFT_ADDR: "0xc6e7DF5E7b4f2A278906862b61205850344D4e7d"
 AFTERMARKET_DEVICE_CONTRACT_ADDRESS: "0xcf9af64522162da85164a714c23a7705e6e466b3"
 SYNTHETIC_DEVICE_CONTRACT_ADDRESS: "0x85226A67FF1b3Ec6cb033162f7df5038a6C3bAB2"
+REWARDS_CONTRACT_ADDRESS: "0x375885164266d48C48abbbb439Be98864Ae62bBE"
```
elffjs, line 17: Newline!

internal/services/contracts_events_consumer.go
```diff
@@ -545,14 +551,16 @@ func (c *ContractsEventsConsumer) handleTokensTransferredForDevice(ctx context.C
 
 	if common.HexToAddress(c.settings.AftermarketDeviceAddr) == args.DeviceNftProxy {
 		reward.AftermarketTokenID = null.IntFrom(int(args.DeviceNode.Int64()))
-		reward.AftermarketEarnings = null.IntFrom(int(args.Amount.Int64()))
+		reward.AftermarketEarnings = types.NewNullDecimal(decimal.New(args.Amount.Int64(), 0))
```
elffjs, line 554: Converting to `int64` is going to be disastrous, I think. You don't want to lose any precision here.

migrations/00021_rewards.sql
```diff
@@ -2,13 +2,13 @@
 -- +goose StatementBegin
 CREATE TABLE rewards (
     issuance_week int NOT NULL,
+    vehicle_id int CONSTRAINT rewards_vehicle_token_id REFERENCES vehicles (id),
     connection_streak int,
-    streak_earnings int,
-    vehicle_id int NOT NULL CONSTRAINT rewards_vehicle_token_id REFERENCES vehicles (id),
+    streak_earnings numeric(30,0),
```
elffjs, line 7: Could you explain where 30 comes from in a comment? I do think it's wide enough, but I'd like to know why.

internal/services/contracts_events_consumer.go
```diff
@@ -153,6 +152,13 @@ func (c *ContractsEventsConsumer) Process(ctx context.Context, event *shared.Clo
 		case VehicleIdChanged:
 			return c.handleVehicleIdChanged(ctx, &data)
 		}
+	case RewardsContractAddr:
```
elffjs, line 155: Good.

elffjs (changes requested):

internal/services/contracts_events_consumer.go
elffjs: I would have a test where the number is so big that it doesn't fit in `int64`.

migrations/00021_rewards.sql
0xdev22: You gave me an example in an earlier comment which was large, about 20. Just gave it a little more allowance in case of something higher.

internal/services/contracts_events_consumer.go
0xdev22: The decimal.new function expects a int64.

migrations/00021_rewards.sql
elffjs: 1 ether = 10^18 wei. The total supply, at the moment, is a billion. So 10^27, with 28 digits, is the biggest number you could possibly get.

elffjs:
My guess is the tests are failing because we never provide the rewards address to the listener.

internal/services/contracts_events_consumer.go
elffjs: Then don't use that function! We have lots of examples of using these types in devices-api and rewards-api—you've experienced it yourself with the privileges work, with the event processing work. See here, for example

https://github.com/DIMO-Network/devices-api/blob/d176e6fc86ac8f69de6688073d007585e44d55e0/internal/services/contracts_events_consumer.go#L214

elffjs: You can also look at https://pkg.go.dev/github.com/ericlagergren/decimal and search for big.Int

elffjs: > I would have a test where the number is so big that it doesn't fit in `int64`.

Did we do this?

charts/identity-api/values.yaml
```diff
@@ -36,6 +36,7 @@ env:
   AFTERMARKET_DEVICE_CONTRACT_ADDRESS: '0xcf9Af64522162da85164a714c23a7705E6e466b3'
   DCN_REGISTRY_ADDR: '0x2daeF6FF0E2B61aaBF9ADDE1aFDAfDb4f0f1E660'
   DCN_RESOLVER_ADDR: '0x7CE3b10571b1ca517dEf45D5F81850467eC8dd2D'
+  REWARDS_CONTRACT_ADDRESS: "0x375885164266d48C48abbbb439Be98864Ae62bBE"
```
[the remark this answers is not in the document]
elffjs: Don't we need the synthetic address, too?

migrations/00021_rewards.sql
elffjs: Could we add that comment?

PR #52, opened 2023-10-25
Si 2149 rewards by vehicle

0xdev22:
- What did you add to the API?
       - New rewards query added to vehicles Query
       - Rewards pagination

- What work does this depend upon?

- What was difficult about this?
    - Handling creation of a SubQuery within the vehicles Query

- What major components did we add?
   - Rewards
   - Breakdown of Earnings for Device, SyntheticDevice and Streak
   

image
_______________________________________________________________________________________

image


graph/schema.graphqls
```diff
@@ -331,3 +333,24 @@ type DCN {
   """
   vehicle: Vehicle
 }
+
+type EarningNode {
+  week: Int!
+  beneficiary: Address!
+  connectionStreak: Int!
+  streakTokens: BigInt!
+  aftermarketDevice: Int!
```
elffjs, line 342: Why an integer and not the type?

Remember, we want the vehicle here too.

graph/schema.graphqls
```diff
@@ -331,3 +333,24 @@ type DCN {
...
+  sentAt: Time!
+}
+
+type EarningTransfers {
+  node: EarningNode!
+}
+
+type VehicleEarnings {
+  earnedTokens: BigInt!
+  earningsTransfers: [EarningTransfers!]!
```
elffjs, line 355: This should be a connection.

graph/types/blockchain.go
```diff
@@ -61,3 +62,18 @@ func MarshalInt(x int) graphql.Marshaler {
...
 var UnmarshalInt = graphql.UnmarshalInt
+
+func MarshalBigInt(x *big.Int) graphql.Marshaler {
+	return graphql.WriterFunc(func(w io.Writer) {
+		_, _ = io.WriteString(w, strconv.Quote(x.String()))
+	})
+}
+
+func UnmarshalBigInt(v interface{}) (*big.Int, error) {
+	s, ok := v.(*big.Int)
```
elffjs, line 73: We should do an end to end test. I think this is going to fail like the `Bytes` stuff failed.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,73 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	earnings := []*gmodel.EarningTransfers{}
+	totalTokensEarned := big.NewInt(0)
+
+	for _, reward := range rewards {
+		stEarn := new(big.Int).SetUint64(toUint64(reward.StreakEarnings))
```
elffjs, line 44: Converting to `uint64` is going to be disastrous.

gqlgen.yml
```diff
@@ -81,15 +81,14 @@ models:
       - github.com/99designs/gqlgen/graphql.Int64
       - github.com/99designs/gqlgen/graphql.Int32
   Int:
-    model:
-      - github.com/99designs/gqlgen/graphql.Int
-      - github.com/99designs/gqlgen/graphql.Int64
-      - github.com/99designs/gqlgen/graphql.Int32
+    model: github.com/DIMO-Network/identity-api/graph/types.Int
```
elffjs, line 84: I think it's dangerous to redefine `Int` to be a string in the output—is that what this will do? I would rather just output a number.

graph/schema.graphqls
```diff
@@ -176,6 +176,8 @@ type Vehicle implements Node {
   definition: Definition
   dcn: DCN
   name: String!
+
+  earnings: VehicleEarnings
```
elffjs, line 180: I like splitting this out into its own sub-object, that's cool.

internal/repositories/rewards.go
elffjs: I would try using `.Int(nil)`

https://pkg.go.dev/github.com/ericlagergren/decimal?utm_source=godoc#Big.Int

But you should check, also, for nils.

I'm going to add some stuff to shared to help with this.

graph/schema.graphqls
```diff
@@ -176,6 +176,8 @@ type Vehicle implements Node {
   definition: Definition
   dcn: DCN
   name: String!
+
+  earnings: VehicleEarningsConnection
```
elffjs, line 180: I guess this shouldn't be the "connection" since it's not the listing.

graph/schema.graphqls
```diff
@@ -331,3 +333,51 @@ type DCN {
...
+  """
+  week: Int!
+  """
+  Address of Beneficiary that received reward
+  """
+  beneficiary: Address!
+  """
+  Consecutive period of which vehicle was connected
+  """
+  connectionStreak: Int!
```
elffjs, line 349: Can you require this? There is a state in which you have device earning information but not streak information.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,55 @@
...
+
+		totalTokensEarned = totalTokensEarned.Add(totalTokensEarned, stEarn)
+		totalTokensEarned = totalTokensEarned.Add(totalTokensEarned, adEarn)
+		totalTokensEarned = totalTokensEarned.Add(totalTokensEarned, syEarn)
+
+		earning := &gmodel.EarningsEdge{
+			Node: &gmodel.EarningNode{
+				Week:                    reward.IssuanceWeek,
+				Beneficiary:             common.BytesToAddress(reward.ReceivedByAddress.Bytes),
+				ConnectionStreak:        reward.ConnectionStreak.Int,
```
elffjs, line 38: Like this `.Int` here makes me nervous.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,55 @@
...
+		return nil, err
+	}
+
+	earnings := []*gmodel.EarningsEdge{}
+	totalTokensEarned := big.NewInt(0)
+
+	for _, reward := range rewards {
+		stEarn := dbtypes.NullDecimalToInt(reward.StreakEarnings)
+		adEarn := dbtypes.NullDecimalToInt(reward.AftermarketEarnings)
+		syEarn := dbtypes.NullDecimalToInt(reward.SyntheticEarnings)
```
elffjs, line 28: I would be very careful. These can be null.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,55 @@
...
+
+		totalTokensEarned = totalTokensEarned.Add(totalTokensEarned, stEarn)
+		totalTokensEarned = totalTokensEarned.Add(totalTokensEarned, adEarn)
+		totalTokensEarned = totalTokensEarned.Add(totalTokensEarned, syEarn)
```
elffjs, line 32: Note that it's the same to write
```
		totalTokensEarned.Add(totalTokensEarned, stEarn)
		totalTokensEarned.Add(totalTokensEarned, adEarn)
		totalTokensEarned.Add(totalTokensEarned, syEarn)
```
It's up to you which one you want.

internal/repositories/rewards.go
elffjs: See here for example:

https://go.dev/play/p/3mZOhSGaFTc

internal/repositories/rewards.go
```diff
@@ -0,0 +1,55 @@
...
+
+		earning := &gmodel.EarningsEdge{
+			Node: &gmodel.EarningNode{
+				Week:                    reward.IssuanceWeek,
+				Beneficiary:             common.BytesToAddress(reward.ReceivedByAddress.Bytes),
+				ConnectionStreak:        reward.ConnectionStreak.Int,
+				StreakTokens:            stEarn,
+				AftermarketDeviceID:     &reward.AftermarketTokenID.Int,
```
elffjs, line 40: I think this is also dangerous. If `AftermarketTokenID` is indeed null then you want this field to also be null. That's going to make your check up [here](https://github.com/DIMO-Network/identity-api/pull/52/files#diff-e0a4909f568588675db7c5cfb3aa4175267381ba017bc73e85f980bd3ed715ffR40) worthwhile.

graph/schema.graphqls
elffjs: You could consider, in your query, hiding rows that don't have a connection streak set. That would be annoying but it would allow you to be very confident that

1. The streak is there
2. At least one of the devices is populated

graph/rewards_test.go
```diff
@@ -0,0 +1,148 @@
...
+	query := `{vehicle(tokenId: 11) {id earnings {earnedTokens earningsTransfers {node {week beneficiary connectionStreak streakTokens aftermarketDevice {id} aftermarketDeviceTokens syntheticDevice {tokenId} syntheticDeviceTokens sentAt}}}}}`
+
+	c := client.New(loader.Middleware(r.pdb, handler.NewDefaultServer(NewExecutableSchema(Config{Resolvers: r.resolver}))))
+
+	var resp interface{}
+	c.MustPost(query, &resp)
+	b, _ := json.Marshal(resp)
+	fmt.Println(string(b))
+
+	r.JSONEq(fmt.Sprintf(`{"vehicle":{"earnings":{"earnedTokens":"177441154036585529047","earningsTransfers":[{"node":{"aftermarketDevice":null,"aftermarketDeviceTokens":"59147051345528509681","beneficiary":"%s","connectionStreak":20,"sentAt":"%s","streakTokens":"59147051345528509684","syntheticDevice":null,"syntheticDeviceTokens":"59147051345528509682","week":1}}]},"id":"V_kQs="}}`, beneficiary.Hex(), currTime.Format(time.RFC3339)), string(b))
```
elffjs, line 147: How can I have `syntheticDeviceTokens` but no `syntheticDevice`? That seems odd.

elffjs: Because we're using `JSONEq` we can format this sanely, by the way.

elffjs:
I think this is moving in a good direction. There are two major hurdles:

1. We have fields in the database that are not required but that we are requiring in the API. I think this is going to lead to problems. Anywhere where you see `.Int` without a guard I think that's suspicious. The ways out are 
   a. Don't require this stuff in the API
   b. Restrict the query to rows where, e.g., the streak is populated

2. We still need to implement all the normal connection stuff.

graph/schema.graphqls
```diff
@@ -331,3 +333,51 @@ type DCN {
...
+  """
+  syntheticDeviceTokens: BigInt!
+  """
+  When the token was earned
+  """
+  sentAt: Time!
+}
+
+type EarningsEdge {
+  node: EarningNode!
```
elffjs, line 377: Cursor?

elffjs:
I think it would also help, by the way, to have a PR description that quickly guides the readier through what he’s looking at.

* What did you add to the API?
* What work does this depend upon?
* What was difficult about this?
* What major components did we add?

internal/loader/aftermarket_device_loader.go
```diff
@@ -54,3 +63,35 @@ func (ad *AftermarketDeviceLoader) BatchGetLinkedAftermarketDeviceByVehicleID(ct
...
+
+	devices, err := models.AftermarketDevices(models.AftermarketDeviceWhere.ID.IN(ids)).All(ctx, ad.db.DBS().Reader)
+	if err != nil {
+		for i := range ids {
+			results[i] = &dataloader.Result[*model.AftermarketDevice]{Data: nil, Error: err}
+		}
+		return results
+	}
+
+	amByVehicleID := map[int]*models.AftermarketDevice{}
```
elffjs, line 80: Why are there all these references to vehicle ids in this function? I'm confused.

I think we should build up a map from aftermarket device id to aftermarket device.

internal/helpers/pagination_helper.go
```diff
@@ -35,3 +37,33 @@ func (p PaginationHelper[T]) DecodeCursor(cursor string) (*T, error) {
...
+
+	if first != nil {
+		if last != nil {
+			return nil, errors.New("pass `first` or `last`, but not both")
+		}
+		if *first < 0 {
+			return nil, errors.New("the value for `first` cannot be negative")
+		}
+		if *first > maxPageSize {
+			return nil, fmt.Errorf("the value %d for `first` exceeds the limit %d", *last, maxPageSize)
```
elffjs, line 52:
```suggestion
			return nil, fmt.Errorf("the value %d for `first` exceeds the limit %d", *first, maxPageSize)
```

internal/loader/aftermarket_device_loader.go
```diff
@@ -54,3 +63,35 @@ func (ad *AftermarketDeviceLoader) BatchGetLinkedAftermarketDeviceByVehicleID(ct
...
+
+	for _, d := range devices {
+		amByVehicleID[d.VehicleID.Int] = d
```
elffjs, line 83: Is there any guarantee that `.Int` is non-zero?

internal/loader/aftermarket_device_loader.go
```diff
@@ -54,3 +63,35 @@ func (ad *AftermarketDeviceLoader) BatchGetLinkedAftermarketDeviceByVehicleID(ct
...
+	}
+
+	for i, vID := range ids {
+		if am, ok := amByVehicleID[vID]; ok {
+			results[i] = &dataloader.Result[*model.AftermarketDevice]{
+				Data: repositories.AftermarketDeviceToAPI(am),
+			}
+		} else {
+			results[i] = &dataloader.Result[*model.AftermarketDevice]{}
```
elffjs, line 92: Maybe this should be an error? We couldn't find it.

internal/loader/synthetic_device_loader.go
```diff
@@ -55,3 +64,36 @@ func (sd *SyntheticDeviceLoader) BatchGetSyntheticDeviceByVehicleID(ctx context.
...
+
+	devices, err := models.SyntheticDevices(models.SyntheticDeviceWhere.ID.IN(ids)).All(ctx, sd.db.DBS().Reader)
+	if err != nil {
+		for i := range ids {
+			results[i] = &dataloader.Result[*gmodel.SyntheticDevice]{Data: nil, Error: err}
+		}
+		return results
+	}
+
+	sdByVehicleID := map[int]*models.SyntheticDevice{}
```
elffjs, line 81: Similarly to aftermarket devices, I don't understand the focus on vehicle ids here.

internal/loader/aftermarket_device_loader.go
elffjs: I think the reason the tests aren't blowing up is that everything has id 1 so it's all kind of working out.

elffjs (changes requested):
More soon. Lots to like here, particularly in the tests, but there do seem to be some gaps.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	var stats rewardStats
+
+	limit := new(int)
+	*limit = 100
+
+	err := models.Rewards(
+		qm.Select(
+			fmt.Sprintf(
+				`sum(%s + %s + %s) as token_sum`,
```
elffjs, line 56: Do you know what happens if there are no rows? We said the total would be `BigInt!`, so it needs to be there. Just checking, since we treat this as `types.NullDecimal`.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,218 @@
...
+
+func RewardToAPI(reward models.Reward) gmodel.Earning {
+	stEarn := dbtypes.NullDecimalToInt(reward.StreakEarnings)
+	adEarn := dbtypes.NullDecimalToInt(reward.AftermarketEarnings)
+	syEarn := dbtypes.NullDecimalToInt(reward.SyntheticEarnings)
+
+	return gmodel.Earning{
+		Week:                    reward.IssuanceWeek,
+		Beneficiary:             common.BytesToAddress(reward.ReceivedByAddress.Bytes),
+		ConnectionStreak:        reward.ConnectionStreak.Ptr(),
```
elffjs, line 30: Good stuff.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func RewardToAPI(reward models.Reward) gmodel.Earning {
+	stEarn := dbtypes.NullDecimalToInt(reward.StreakEarnings)
+	adEarn := dbtypes.NullDecimalToInt(reward.AftermarketEarnings)
+	syEarn := dbtypes.NullDecimalToInt(reward.SyntheticEarnings)
+
+	return gmodel.Earning{
+		Week:                    reward.IssuanceWeek,
+		Beneficiary:             common.BytesToAddress(reward.ReceivedByAddress.Bytes),
```
elffjs, line 29: This `.Bytes` weirds me out. Is this not required in the schema?

internal/repositories/rewards.go
```diff
@@ -0,0 +1,218 @@
...
+		return nil, err
+	}
+
+	if vehicleEarnings == nil || vehicleEarnings.History == nil {
+		return &gmodel.EarningsConnection{}, nil
+	}
+
+	erningsConnectionCopy := *vehicleEarnings.History
+	earningIdx := map[string]int{}
+	for idx, val := range erningsConnectionCopy.Edges {
```
elffjs, line 149: I don't think this interplay between cursors and limits is going to work out. We took the limit in the parent resolver before ever seeing the cursor. Like suppose our list is

* Week 4
* Week 3
* Week 2
* Week 1

We get arguments of `first: 2` and `after: week 3`. In the parent resolver I think we're going to slap a `LIMT 2` on the query and get

* Week 4
* Week 3

But the request calls for 

* Week 2
* Week 1

How will this work out?

internal/repositories/rewards.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if vehicleEarnings == nil || vehicleEarnings.History == nil {
+		return &gmodel.EarningsConnection{}, nil
+	}
+
+	erningsConnectionCopy := *vehicleEarnings.History
+	earningIdx := map[string]int{}
```
elffjs, line 148: Okay, so we're trying to build a map from cursors to slots.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,218 @@
...
+				PageInfo:   &gmodel.PageInfo{},
+				TotalCount: vehicleEarnings.History.TotalCount,
+			}, nil
+		}
+
+		erningsConnectionCopy.Edges = erningsConnectionCopy.Edges[:startFrom]
+		erningsConnectionCopy.Nodes = erningsConnectionCopy.Nodes[:startFrom]
+	}
+
+	if before != nil || last != nil {
```
elffjs, line 188: I don't think `before != nil` implies that you should reverse.

elffjs: But this is another problem—in the parent where we did the select, did we ever reverse the order in the case where `last` was specified?

internal/repositories/rewards.go
```diff
@@ -0,0 +1,218 @@
...
+		),
+		models.RewardWhere.VehicleID.EQ(tokenID),
+	).Bind(ctx, r.pdb.DBS().Reader, &stats)
+	if err != nil {
+		return nil, err
+	}
+
+	queryMods := []qm.QueryMod{
+		models.RewardWhere.VehicleID.EQ(tokenID),
+		qm.OrderBy(models.RewardColumns.IssuanceWeek + " desc"),
```
elffjs, line 69: Is it concerning that this is always `DESC`?

internal/repositories/rewards.go
```diff
@@ -0,0 +1,218 @@
...
+
+	if before != nil || last != nil {
+		slices.Reverse(erningsConnectionCopy.Edges)
+		slices.Reverse(erningsConnectionCopy.Nodes)
+	}
+
+	if *limit > len(erningsConnectionCopy.Edges) { // out of bounds protection
+		*limit = len(erningsConnectionCopy.Edges)
+	}
+	edges = erningsConnectionCopy.Edges[:*limit]
```
elffjs, line 196: I think we should probably do this before reversing?

internal/repositories/rewards.go
```diff
@@ -0,0 +1,218 @@
...
+	limit, err := pHelp.ValidateFirstLast(first, last, maxPageSize)
+	if err != nil {
+		return nil, err
+	}
+
+	if vehicleEarnings == nil || vehicleEarnings.History == nil {
+		return &gmodel.EarningsConnection{}, nil
+	}
+
+	erningsConnectionCopy := *vehicleEarnings.History
```
elffjs, line 147: This copy stuff makes it very hard to follow this function.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	queryMods := []qm.QueryMod{
+		models.RewardWhere.VehicleID.EQ(tokenID),
+		qm.OrderBy(models.RewardColumns.IssuanceWeek + " desc"),
+	}
+
+	queryMods = append(queryMods, qm.Limit(*limit))
```
elffjs, line 72: Should this be `+1?

gqlgen.yml
```diff
@@ -89,6 +89,7 @@ models:
     model: github.com/DIMO-Network/identity-api/graph/types.Address
   BigInt:
     model:
+      - github.com/DIMO-Network/identity-api/graph/types.BigInt
       - github.com/DIMO-Network/identity-api/graph/types.Int
```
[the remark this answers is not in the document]
0xdev22: What line, the Int ?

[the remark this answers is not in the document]
elffjs: Yes. We're not using it.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func (r *Repository) PaginateVehicleEarningsByID(ctx context.Context, vehicleEarnings *gmodel.VehicleEarnings, first *int, after *string, last *int, before *string) (*gmodel.EarningsConnection, error) {
+	pHelp := helpers.PaginationHelper[RewardsCursor]{}
+	limit, err := pHelp.ValidateFirstLast(first, last, maxPageSize)
+	if err != nil {
+		return nil, err
+	}
+
+	if vehicleEarnings == nil || vehicleEarnings.History == nil {
```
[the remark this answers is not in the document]
0xdev22: .History is not nil. It already contains the history from [here](https://github.com/DIMO-Network/identity-api/pull/52/files/e3a1b03b51a0c75df11c58002ac9348d5b121894#diff-ee2cbc68675eedede463a291d110675efe40445447c9a282b3386a725eb44688R41). For all rewards for the specific tokenID. I explained this via discord the other day.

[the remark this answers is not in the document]
elffjs: I know it's not nil because you populated it, but I think that's a bad idea.

internal/helpers/pagination_helper.go
```diff
@@ -35,3 +37,33 @@ func (p PaginationHelper[T]) DecodeCursor(cursor string) (*T, error) {
 
 	return &res, nil
 }
+
+func (p PaginationHelper[T]) ValidateFirstLast(first, last *int, maxPageSize int) (*int, error) {
```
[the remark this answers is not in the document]
0xdev22: You are right, will remove it as part of the Pagination helper, it can be a standalone function in thesame file. However, pagination helper has a generic type, because it does just that, provides a generic function to help return a cursor of any type.

internal/repositories/rewards.go
0xdev22: It returns `nil`

elffjs:
I wonder if some of these are failing on main as well. I'll take a look.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,174 @@
...
+		SyntheticDeviceTokens:   syEarn,
+		SentAt:                  reward.EarnedAt,
+		VehicleID:               reward.VehicleID,
+	}
+}
+
+func (r *Repository) GetEarningsByVehicleID(ctx context.Context, tokenID int) (*gmodel.VehicleEarnings, error) {
+	type rewardStats struct {
+		TokenSum   types.NullDecimal `boil:"token_sum"`
+		TotalCount int               `boil:"total_count"`
```
elffjs, line 44: My opinion is that it would be better to do this in the child resolver. It's a bit out of place here and I worry that if we introduce any filtering or anything it's gonna get lost.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	var stats rewardStats
+
+	limit := new(int)
+	*limit = 100
```
elffjs, line 49: Remove?

internal/repositories/rewards.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	queryMods := []qm.QueryMod{
+		models.RewardWhere.VehicleID.EQ(vehicleEarnings.VehicleID),
+	}
+
+	if after != nil {
+		afterID, err := helpers.CursorToID(*after)
```
elffjs, line 95: It looks like the cursors are the weeks now. This is fine for by-vehicle and by-device, but we will probably have to do something different when we look at by-user.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,174 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		queryMods = append(queryMods, models.RewardWhere.IssuanceWeek.GT(beforeID))
+	}
+
+	orderBy := "DESC"
+	if last != nil {
+		orderBy = "ASC"
```
elffjs, line 114: We probably need to reverse later, yeah?

Lord help us when we introduce sort options.

internal/repositories/rewards_test.go
```diff
@@ -185,7 +185,7 @@ func (r *RewardsRepoTestSuite) Test_GetEarningsByVehicleID_NoRows() {
 	r.NoError(err)
 
 	r.Equal(&gmodel.VehicleEarnings{
-		TotalTokens: nil,
+		TotalTokens: &big.Int{},
```
elffjs, line 188: Maybe `big.NewInt(0)`?

internal/repositories/vehicles_test.go
```diff
@@ -785,37 +785,37 @@ func (o *AccessibleVehiclesRepoTestSuite) Test_GetAccessibleVehicles_Pagination_
 
 	last := 2
 	after := "NA=="
-	startCrsr := "Mw=="
+	startCrsr := "MQ=="
```
elffjs, line 788: If this comes from fixing pagination, can you cherry-pick all of this to main?

internal/repositories/vehicles_test.go
```diff
@@ -785,37 +785,37 @@ func (o *AccessibleVehiclesRepoTestSuite) Test_GetAccessibleVehicles_Pagination_
...
 		HasPreviousPage: true,
 		HasNextPage:     false,
-	})
+	}, res.PageInfo)
 	expected := []*gmodel.VehicleEdge{
 		{
 			Node: &gmodel.Vehicle{
-				ID:                "V_kQM=",
-				TokenID:           3,
+				ID:                "V_kQE=",
```
elffjs, line 804: To be honest, does this seem right at all? I feel like we're not flipping it.

0xdev22: Tbh, since this isn't within the scope of this task, I fixed it to allow the tests pass, didn't observe the behaviour as much. Since the last change that broke the tests, I think that was when you used last to determine if we apply `ASC`, I'm not sure of what the current behaviour is.

elffjs: Merge in main and it should work.

elffjs: Oh, I did that actually.

internal/repositories/rewards.go
0xdev22: This is the bulk of what is being done in this parent resolver. Wouldn't that render the parent resolver redundant and unecessary.

internal/repositories/rewards.go
0xdev22: reverse ?

internal/repositories/vehicles_test.go
0xdev22: Sure

internal/repositories/rewards.go
elffjs: Is this resolved?

internal/repositories/rewards.go
elffjs: Did you get a chance to check?

internal/repositories/rewards.go
elffjs: But clients are expecting to receive these by week, descending. Specifying `first` or `last` is not supposed to change the order, just the records that are returned.

elffjs: Look at vehicles and aftermarket devices for a guide.

internal/loader/synthetic_device_loader.go
```diff
@@ -55,3 +65,23 @@ func (sd *SyntheticDeviceLoader) BatchGetSyntheticDeviceByVehicleID(ctx context.
...
+	if err != nil {
+		for i := range results {
+			results[i] = &dataloader.Result[*model.SyntheticDevice]{Error: err}
+		}
+		return results
+	}
+
+	for i, sdv := range devices {
+		results[i] = &dataloader.Result[*gmodel.SyntheticDevice]{Data: repositories.SyntheticDeviceToAPI(sdv)}
+	}
```
elffjs, line 84: I think you still have to build up a map and be careful about order. You do not know whether

* `syntheticDeviceIDs` is without repeats
* the items are coming back in the order that you want

internal/loader/aftermarket_device_loader.go
```diff
@@ -54,3 +63,23 @@ func (ad *AftermarketDeviceLoader) BatchGetLinkedAftermarketDeviceByVehicleID(ct
...
+	if err != nil {
+		for i := range ids {
+			results[i] = &dataloader.Result[*model.AftermarketDevice]{Error: err}
+		}
+		return results
+	}
+
+	for i, adv := range devices {
+		results[i] = &dataloader.Result[*model.AftermarketDevice]{Data: repositories.AftermarketDeviceToAPI(adv)}
+	}
```
elffjs, line 82: Similar comment here.

internal/repositories/rewards.go
elffjs: I guess I don't worry about whether a resolver has been given sufficient meaning in life. Here the resolver being invoked means we dip into the rewards table or not, that seems like enough distinction to me.

Anyway, we can leave it the way it is.

internal/repositories/rewards.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil, err
+	}
+
+	if stats.TokenSum.IsZero() {
+		return &gmodel.VehicleEarnings{
+			TotalTokens: &big.Int{},
+			History:     &gmodel.EarningsConnection{},
+			VehicleID:   tokenID,
+		}, nil
+	}
```
elffjs, line 69: Does this do much of anything?

elffjs (approved):

elffjs:
@Allyson-English Take a look?

Allyson-English (approved):

PR #56, opened 2023-11-06
SI-2159-filter-AD-by-beneficiary

Allyson-English:
[ticket](https://linear.app/dimo/issue/SI-2159/filter-aftermarket-devices-by-beneficiary)

- Allow clients to filter the aftermarket devices list by beneficiary

### DIFF
--- a/internal/repositories/aftermarket_devices_test.go
+++ b/internal/repositories/aftermarket_devices_test.go
@@ -1,9 +1,7 @@
 package repositories
 
 import (
-	"bytes"
 	"context"
-	"encoding/base64"
 	"fmt"
 	"math/big"
 	"strings"
@@ -16,7 +14,6 @@ import (
 	"github.com/DIMO-Network/identity-api/models"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/stretchr/testify/assert"
-	"github.com/vmihailenco/msgpack/v5"
 	"github.com/volatiletech/sqlboiler/v4/boil"
 )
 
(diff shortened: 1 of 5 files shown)


internal/repositories/aftermarket_devices_test.go
```diff
@@ -134,3 +137,60 @@ func Test_GetOwnedAftermarketDevices_Pagination_PreviousPage(t *testing.T) {
...
+	e := msgpack.NewEncoder(&b)
+	e.UseArrayEncodedStructs(true)
+	_ = e.Encode(aftermarketDevicePrimaryKey{TokenID: tknID})
+	id := "AD_" + base64.StdEncoding.EncodeToString(b.Bytes())
+	mn, err := gmn.EntropyToMnemonicThreeWords(aftermarketDeviceNodeMintedArgs.AftermarketDeviceAddress.Bytes())
+	if err != nil {
+		assert.NoError(t, err)
+	}
+	name := strings.Join(mn, " ")
+	expectedOwnerFilterResp := []*model.AftermarketDeviceEdge{
```
elffjs, line 172: No way could this pass, yeah? You're going to get all 4.

elffjs: I'm stupid.

elffjs:
@0xdev22 Can you take a look?

internal/repositories/aftermarket_devices_test.go
```diff
@@ -134,3 +134,54 @@ func Test_GetOwnedAftermarketDevices_Pagination_PreviousPage(t *testing.T) {
 	}
 	assert.Exactly(t, expected, res.Edges)
 }
+
+func Test_GetAftermarketDevices_FilterByBeneficiary(t *testing.T) {
```
0xdev22, line 138: Some missing tests

- When beneficiary has no records (**another beneficary's id is supplied that has no records**)
- Both Owner and Beneficiary are supplied

0xdev22 (changes requested):

internal/repositories/aftermarket_devices_test.go
elffjs: I'm okay to go without these given how orthogonal the code is with everything else. Let's focus on tests for the other PR, which is a bit more intricate.

PR #59, opened 2023-11-07
SI-2166-filter-vehicles-by-owner

Allyson-English:
[ticket](https://linear.app/dimo/issue/SI-2166/filter-vehicles-by-owner-not-just-privileged)

- add ability to filter vehicles by owner 

### Notes
- spent a lot of time thinking about how we could get rid of the big if/ else statement for querying by privileges versus all vehicles/ by owner but didn't think of anything better than what's there (thought for a while we could use a WITH query in a creative way) open to ideas though!


internal/repositories/vehicles.go
```diff
@@ -137,31 +137,31 @@ func (v *Repository) GetVehicles(ctx context.Context, first *int, after *string,
...
-			return nil, errors.New("Provide `first` or `last`.")
+			return nil, errors.New("provide `first` or `last`")
 		}
 		if *last < 0 {
-			return nil, errors.New("The value for `last` cannot be negative.")
+			return nil, errors.New("the value for `last` cannot be negative")
 		}
 		if *last > maxPageSize {
-			return nil, fmt.Errorf("The value %d for `last` exceeds the limit %d.", *last, maxPageSize)
+			return nil, fmt.Errorf("the value %d for `last` exceeds the limit %d", *last, maxPageSize)
```
elffjs, line 157: We actually want this.

elffjs: Let me just figure out how to make VS Code ignore this. The problem is that we want nice sentences for the client but Go errors aren't supposed to have these. golangci-lint does not have a problem with it.

internal/repositories/vehicles.go
```diff
@@ -137,31 +137,31 @@ func (v *Repository) GetVehicles(ctx context.Context, first *int, after *string,
...
 		}
 		limit = *last
 	}
 
+	var err error
 	var totalCount int64
 	var queryMods []qm.QueryMod
-
 	if filterBy != nil && filterBy.Privileged != nil {
```
elffjs, line 165: What if someone places both filters on?

internal/repositories/vehicles.go
```diff
@@ -191,15 +190,18 @@ func (v *Repository) GetVehicles(ctx context.Context, first *int, after *string,
 			return nil, err
 		}
 	} else {
-		// TODO(elffjs): Ugly.
-		queryMods = []qm.QueryMod{}
```
elffjs, line 195: I think this is here for a reason? Hm, maybe not

https://go.dev/play/p/76sz0-OLykg

internal/repositories/vehicles.go
```diff
@@ -137,31 +137,31 @@ func (v *Repository) GetVehicles(ctx context.Context, first *int, after *string,
...
+			return nil, errors.New("the value for `last` cannot be negative")
 		}
 		if *last > maxPageSize {
-			return nil, fmt.Errorf("The value %d for `last` exceeds the limit %d.", *last, maxPageSize)
+			return nil, fmt.Errorf("the value %d for `last` exceeds the limit %d", *last, maxPageSize)
 		}
 		limit = *last
 	}
 
+	var err error
```
elffjs, line 162: I honestly don't remember why these are all declared up here.

internal/repositories/owned_vehicles_test.go
```diff
@@ -290,3 +291,101 @@ func (s *OwnedVehiclesRepoTestSuite) Test_GetOwnedVehicles_Pagination_NextPage()
...
+	}
+
+	for _, p := range privTbl {
+		if err := p.Insert(ctx, pdb.DBS().Writer, boil.Infer()); err != nil {
+			assert.NoError(err)
+		}
+	}
+
+	if a, err := models.Privileges().All(ctx, pdb.DBS().Reader); err != nil {
+		assert.NoError(err)
```
elffjs, line 348: Well, this is a funny thing to write.

internal/repositories/vehicles.go
Allyson-English: I think that case is already accounted for in the privileges query because they inherently have privileges on vehicles they own, this is the portion of the privileges query that seems to accomplish that:

```
	qm.Expr(
		models.VehicleWhere.OwnerAddress.EQ(addr.Bytes()),
		qm.Or2(
			qm.Expr(
				models.PrivilegeWhere.UserAddress.EQ(addr.Bytes()),
				models.PrivilegeWhere.ExpiresAt.GTE(time.Now()),
			),
		),
	),
```

elffjs: Suppose we have

```
Vehicle | Owner | Privileged Users
--------+-------+-----------------
1       | A     | B
2       | B     | C
```

If I ask for vehicles which `A` owns and to which `B` has privileges, I'm going to get _both_ of these. I guess the fundamental question is whether you think this is an OR or an AND. I think it's an AND.

internal/repositories/owned_vehicles_test.go
```diff
@@ -290,3 +291,100 @@ func (s *OwnedVehiclesRepoTestSuite) Test_GetOwnedVehicles_Pagination_NextPage()
 
 	s.Exactly(expected, res.Edges)
 }
+
+func Test_GetOwnedVehicles_Filters(t *testing.T) {
```
0xdev22, line 295: We should have more tests, maybe testing the combinations mentioned by @elffjs 

- Filter by Owner (Done already)
- Filter by both owner and beneficiary
- Filter by invalid beneficiary, but valid owner
- Filter by invalid owner, but valid beneficiary
- Filter by invalid beneficiary and owner

0xdev22 (changes requested):

internal/repositories/owned_vehicles_test.go
```diff
@@ -295,3 +296,119 @@ func (s *OwnedVehiclesRepoTestSuite) Test_GetOwnedVehicles_Pagination_NextPage()
...
+			}
+		}
+	}
+
+	// Filter By Privileged
+	// Where Priv wallet has both implied (owner) and explicit (granted) privileges
+	// Owner | Privileged | Result
+	// ------+------------+-------
+	// 	     | B          | 1, 2
+	res, err := repo.GetVehicles(ctx, &first, nil, nil, nil, &gmodel.VehiclesFilter{Privileged: walletB})
```
elffjs, line 371: Could we organize these as sub-tests?

internal/repositories/owned_vehicles_test.go
```diff
@@ -295,3 +296,119 @@ func (s *OwnedVehiclesRepoTestSuite) Test_GetOwnedVehicles_Pagination_NextPage()
...
+				assert.NoError(err)
+			}
+		}
+	}
+
+	// Filter By Privileged
+	// Where Priv wallet has both implied (owner) and explicit (granted) privileges
+	// Owner | Privileged | Result
+	// ------+------------+-------
+	// 	     | B          | 1, 2
```
elffjs, line 370: Alignment seems off on some of these.

PR #62, opened 2023-11-09
Rewards integration with AftermarketDevice

0xdev22:


graph/schema.graphqls
```diff
@@ -274,6 +279,15 @@ type AftermarketDevice implements Node {
   The Image Url of the device
   """
   image: String!
+  """
+  The earnings attached to the aftermarket device
+  """
+  earnings(
```
elffjs, line 285: We seem to have gone against what we did for `Vehicle`, where the "big" resolver is one level deeper.

graph/schema.graphqls
```diff
@@ -219,6 +219,11 @@ type VehicleEarnings {
   ): EarningsConnection!
 }
 
+type AfterMarketEarnings {
```
elffjs, line 222:
```suggestion
type AftermarketDeviceEarnings {
```

internal/repositories/rewards.go
```diff
@@ -39,55 +40,10 @@ func RewardToAPI(reward models.Reward) gmodel.Earning {
...
-		VehicleID:   tokenID,
-	}, nil
-}
-
-func (r *Repository) PaginateVehicleEarningsByID(ctx context.Context, vehicleEarnings *gmodel.VehicleEarnings, first *int, after *string, last *int, before *string) (*gmodel.EarningsConnection, error) {
-	limit, err := helpers.ValidateFirstLast(first, last, maxPageSize)
-	if err != nil {
-		return nil, err
-	}
+func (r *Repository) paginateRewards(ctx context.Context, condition []qm.QueryMod, first *int, after *string, last *int, before *string, limit int) (*gmodel.EarningsConnection, error) {
```
elffjs, line 43: `conditions` plural?

graph/schema.graphqls
0xdev22: I felt since this works with a single resolver, it makes things simpler, rather than using the multiple resolvers like we had in the vehicle endpoint.

elffjs: Doing it this way breaks from the Relay spec and makes our API non-uniform. I don't think a slightly simpler implementation is a strong enough argument in the other direction.

graph/rewards_test.go
```diff
@@ -866,3 +866,123 @@ func (r *RewardsQueryTestSuite) Test_Query_GetEarningsByVehicle_BackPaginate_Las
...
+func (r *RewardsQueryTestSuite) Test_Query_GetAftermarketDeviceEarnings_FwdPaginate() {
+	currTime := time.Now().UTC().Truncate(time.Second)
+	_, beneficiary, err := test.GenerateWallet()
+	r.NoError(err)
+
+	r.createDependencies()
+
+	// Aftermarket Earnings
+	adEarn, ok := new(big.Int).SetString("59147051345528509681", 10)
+	r.NotZero(ok)
```
elffjs, line 879: This seems more natural:

https://pkg.go.dev/github.com/stretchr/testify/assert#True

elffjs: I do love these E2E tests.

internal/repositories/rewards.go
```diff
@@ -163,14 +119,136 @@ func (r *Repository) PaginateVehicleEarningsByID(ctx context.Context, vehicleEar
 		startCursor = &sc
 	}
```
Allyson-English, line 120: could references edges here:


	if len(all) != 0 {
		startCur = &edges[0].Cursor
		endCur = &edges[len(edges)-1].Cursor
	}

internal/repositories/rewards.go
```diff
@@ -163,14 +119,136 @@ func (r *Repository) PaginateVehicleEarningsByID(ctx context.Context, vehicleEar
 		startCursor = &sc
```
Allyson-English, line 119: could we reference the edges array made above here?
	var endCur, startCur *string

	if len(all) != 0 {
		startCur = &edges[0].Cursor
		endCur = &edges[len(edges)-1].Cursor
	}

elffjs:
Tests are passing locally. Not sure what's going on with the GitHub runner.

internal/repositories/rewards.go
```diff
@@ -163,14 +119,136 @@ func (r *Repository) PaginateVehicleEarningsByID(ctx context.Context, vehicleEar
...
+	}
+
+	return &gmodel.AftermarketDeviceEarnings{
+		TotalTokens:         dbtypes.NullDecimalToInt(stats.TokenSum),
+		History:             earningsConn,
+		AftermarketDeviceID: tokenID,
+	}, nil
+}
+
+func (r *Repository) PaginateAfterMarketDeviceEarningsByID(ctx context.Context, afterMarketDeviceEarnings *gmodel.AftermarketDeviceEarnings, first *int, after *string, last *int, before *string) (*gmodel.EarningsConnection, error) {
```
elffjs, line 235:
```suggestion
func (r *Repository) PaginateAftermarketDeviceEarningsByID(ctx context.Context, aftermarketDeviceEarnings *gmodel.AftermarketDeviceEarnings, first *int, after *string, last *int, before *string) (*gmodel.EarningsConnection, error) {
```

elffjs (approved):

PR #65, opened 2023-11-21
Change reward synthetic device token to null when syntheticDevice is getting nulled

0xdev22:


internal/services/contracts_events_consumer.go
```diff
@@ -529,7 +529,19 @@ func (c *ContractsEventsConsumer) handleSyntheticDeviceNodeBurnedEvent(ctx conte
 		ID: int(args.SyntheticDeviceNode.Int64()),
 	}
 
-	_, err := sd.Delete(ctx, c.dbs.DBS().Writer)
+	rw, err := models.Rewards(models.RewardWhere.SyntheticTokenID.EQ(null.IntFrom(sd.ID))).One(ctx, c.dbs.DBS().Reader)
```
elffjs, line 532: What do you think about adding an `ON DELETE CASCADE` to the schema? The test wouldn't need to change. Is it better or worse?

elffjs: Whoops, sorry, corrected—meant set null. Cascade is bad!

PR #72, opened 2024-01-08
SI-2237-query-for-manufacturers

Allyson-English:
- Add ability to query for manufacturers and traverse down to the devices one can create a graph visualization dashboard

### DIFF
--- a/internal/repositories/aftermarket_devices.go
+++ b/internal/repositories/aftermarket_devices.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/DIMO-Network/identity-api/graph/model"
 	gmodel "github.com/DIMO-Network/identity-api/graph/model"
 	"github.com/DIMO-Network/identity-api/internal/helpers"
 	"github.com/DIMO-Network/identity-api/models"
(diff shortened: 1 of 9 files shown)


internal/repositories/aftermarket_devices.go
```diff
@@ -225,3 +226,20 @@ func countTrue(ps ...bool) int {
 
 	return out
 }
+
+func (r *Repository) GetAftermarketDevicesForManufacturer(ctx context.Context, obj *model.Manufacturer, first *int, after *string, last *int, before *string, filterBy *model.AftermarketDevicesFilter) (*gmodel.AftermarketDeviceConnection, error) {
+	if filterBy != nil {
+		if filterBy.ManufacturerID != nil {
+			if filterBy.ManufacturerID != &obj.TokenID {
+				return nil, gqlerror.Errorf("Aftermarket device filter must be consistent with manufacturer query.")
+			}
```
elffjs, line 235: This seems awkward. I would rather just not have this as an option.

elffjs:
Could we do a branch where it's not possible to filter by manufacturer within a manufacturer? Just see how ugly that is in code?

I just don't like giving people an API that they can hold wrong if I can help it. I grant that we're already doing this `@oneOf` thing that they can certainly hold wrong. Alas.

PR #76, opened 2024-01-29
[SM-464] add serial filter for aftermarket devices

Asfiroth:


elffjs:
If you're filtering by serial then you're filtering for a single device, no? We already have `aftermarketDevice(by: {serial: "X"}) { … }`. I don't see why we would add this.

PR #88, opened 2024-04-05
SI-2544 - Feature/identity api devicedefinition query tableland

rhvivancoeffio:


internal/config/settings.go
```diff
@@ -19,5 +19,6 @@ type Settings struct {
 	RewardsContractAddr   string      `yaml:"REWARDS_CONTRACT_ADDRESS"`
 	BaseImageURL          string      `yaml:"BASE_IMAGE_URL"`
 
-	BaseVehicleDataURI string `yaml:"BASE_VEHICLE_DATA_URI"`
+	BaseVehicleDataURI  string `yaml:"BASE_VEHICLE_DATA_URI"`
+	TablelandAPIGateway string `yaml:"TABLELAND_API_GATEWAY"`
```
KevinJoiner, line 23: We will also need to add this variable to the values.yaml and values-prod.yaml file under the `charts` dir

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+	"fmt"
+	gmodel "github.com/DIMO-Network/identity-api/graph/model"
+	"github.com/DIMO-Network/identity-api/internal/helpers"
+	"github.com/DIMO-Network/identity-api/internal/repositories/base"
+	"github.com/vektah/gqlparser/v2/gqlerror"
+	"golang.org/x/exp/slices"
+	"net/http"
+	"net/url"
+	"path"
+	"strings"
```
KevinJoiner, line 15: std lib vars should be separate from external imports if you are using vscode this can be done by changing the golang formater to` goimports`

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+	KSUID    string `json:"ksuid"`
+	Model    string `json:"model"`
+	Year     int    `json:"year"`
+	Metadata struct {
+		DeviceAttributes []struct {
+			Name  string `json:"name"`
+			Value string `json:"value,omitempty"`
+		} `json:"device_attributes"`
+	} `json:"metadata"`
+}
```
KevinJoiner, line 224: Move struct definitions to the top of the file.

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+
+func ToAPI(v *DeviceDefinitionTablelandModel) *gmodel.DeviceDefinition {
+	var result = gmodel.DeviceDefinition{}
+
+	result.ID = v.ID
+	result.Ksuid = &v.KSUID
+	result.Year = &v.Year
+	result.Model = &v.Model
+
+	if v.Metadata.DeviceAttributes != nil && len(v.Metadata.DeviceAttributes) > 0 {
```
KevinJoiner, line 33: You do not need this check; in Golang, it is safe to range over a nil or empty list.
```suggestion
```

graph/schema/devicedefinition.graphqls
```diff
@@ -0,0 +1,160 @@
...
+"""
+input DevicedefinitionBy @oneOf {
+  """
+  The manufacturer for the device definition.
+  """
+  manufacturer: String
+  """
+  The id for the device definition.
+  """
+  id: String
```
KevinJoiner, line 133: using `!` means the field is required
```suggestion
  manufacturer: String!
  """
  The id for the device definition.
  """
  id: String!
```

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+func (r *DeviceDefinitionRepository) GetDeviceDefinition(ctx context.Context, by gmodel.DevicedefinitionBy) (*gmodel.DeviceDefinition, error) {
+	if by.Manufacturer == nil || *by.Manufacturer == "" {
+		return nil, gqlerror.Errorf("Provide exactly one `manufacturer`.")
+	}
+
+	if by.ID == nil || *by.ID == "" {
+		return nil, gqlerror.Errorf("Provide exactly one `ID`.")
+	}
+
+	tableName := "_80001_8439"
```
KevinJoiner, line 54: This can be a static const defined at the top of the file.

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+	statement := fmt.Sprintf("SELECT * FROM %s WHERE id = '%s'", tableName, *by.ID)
```
KevinJoiner, line 55: Do you know how tableland handles query injection because right now this is unsafe because the value for by.ID could be `"1; DROP TABLES;`

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if filterBy.Manufacturer == nil || *filterBy.Manufacturer == "" {
+		return nil, gqlerror.Errorf("Provide exactly one `manufacturer`.")
+	}
+
+	var conditions []string
+	if filterBy.Year != nil && (*filterBy.Year > 1980 && *filterBy.Year < 2999) {
```
KevinJoiner, line 83: We should return an error if the year is out of range.

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+		conditions = append(conditions, fmt.Sprintf("year = %d", *filterBy.Year))
+	}
+	if filterBy.Model != nil && len(*filterBy.Model) > 0 {
```
KevinJoiner, line 86: Use empty string check over 0 len for strings.
Also we should return an error if an empty string is given because that means the users explicitly set the value but set it to `""`
```suggestion
	if filterBy.Model != nil && *filterBy.Model != "" {
```

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+		conditions = append(conditions, fmt.Sprintf("model = '%s'", *filterBy.Model))
+	}
+	if filterBy.ID != nil && len(*filterBy.ID) > 0 {
```
KevinJoiner, line 89:
```suggestion
	if filterBy.ID != nil && *filterBy.ID != "" {
```

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+		conditions = append(conditions, fmt.Sprintf("id = '%s'", *filterBy.ID))
+	}
+
+	whereClause := strings.Join(conditions, " AND ")
+	if whereClause != "" {
+		whereClause = " WHERE " + whereClause
+	}
+
+	tableName := "_80001_8439"
```
KevinJoiner, line 98: can be a package level const

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+const TokenPrefix = "DD"
+
+type DeviceDefinitionRepository struct {
+	*base.Repository
+}
+
+func ToAPI(v *DeviceDefinitionTablelandModel) *gmodel.DeviceDefinition {
+	var result = gmodel.DeviceDefinition{}
+
+	result.ID = v.ID
```
KevinJoiner, line 28: These IDs do not have the same meaning as the model.DeviceDefinition.ID is an ID specific to the identity-API that allows users to look up the Query for the DeviceDefinition using the Node query 
as seen here https://github.com/DIMO-Network/identity-api/blob/4aac3c1fedec102555242b93195aeba37fca5b05/graph/schema.resolvers.go#L23-L43

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+	return res, nil
+}
+
+func (r *DeviceDefinitionRepository) QueryTableland(queryParams map[string]string, result interface{}) error {
+	fullURL, err := url.Parse(r.Settings.TablelandAPIGateway)
+	if err != nil {
+		return err
+	}
+
+	fullURL.Path = path.Join(fullURL.Path, "api/v1/query")
```
KevinJoiner, line 186: Use this Join Path function it is safer https://pkg.go.dev/net/url#URL.JoinPath
```suggestion
	fullURL = fullURL.JoinPath(fullURL.Path, "api/v1/query")
```

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+
+	if queryParams != nil {
+		values := fullURL.Query()
+		for key, value := range queryParams {
+			values.Set(key, value)
+		}
+		fullURL.RawQuery = values.Encode()
+	}
+
+	resp, err := http.Get(fullURL.String())
```
KevinJoiner, line 196: We want to call this request with a context because if we don't this could block forever.
```suggestion
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fullURL.String(), nil)
	if err != nil {
		return nil, fmt.Errorf("failed to create request: %w", err)
	}

	resp, err := http.DefaultClient.Do(req)
	if err != nil {
		return nil, fmt.Errorf("failed to complete request: %w", err)
	}
```

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,224 @@
...
+
+type DeviceDefinitionTablelandModel struct {
+	ID       string `json:"id"`
+	KSUID    string `json:"ksuid"`
+	Model    string `json:"model"`
+	Year     int    `json:"year"`
+	Metadata struct {
+		DeviceAttributes []struct {
+			Name  string `json:"name"`
+			Value string `json:"value,omitempty"`
```
KevinJoiner, line 221: nit: Do we need omitempty on this field only?

graph/schema/devicedefinition.graphqls
```diff
@@ -0,0 +1,160 @@
...
+
+"""
+Represents a Device Definition.
+"""
+type DeviceDefinition implements Node {
+  """
+  An opaque global identifier for this device definition.
+  """
+  id: ID!
+
```
KevinJoiner, line 51: Need a separate ID that corresponds to the tableland row
```suggestion
DefinitionID: ID!
```

internal/repositories/devicedefinition/devicedefinition.go
KevinJoiner, line 28:
```suggestion
    globalID, err := base.EncodeGlobalTokenID(TokenPrefix, v.ID)
	if err != nil {
		return nil, fmt.Errorf("error encoding device definition id: %w", err)
	}

	result.ID = globalID
	result.DefinitionID = v.ID
```

KevinJoiner (changes requested):
We might want to change the name from device definition to vehicle definitions since these queries only relate to vehicles and would not make sense for other devices in the future.

.gitignore
```diff
@@ -23,3 +23,5 @@ settings.yaml
 
 .idea
 
+# Desktop Service Store (macOS)
+.DS_Store
```
elffjs, line 27: Can we do a newline at the end of these ppc

cmd/identity-api/main.go
```diff
@@ -78,19 +74,19 @@ func main() {
...
-	}
-
-	cevConsumer := services.NewContractsEventsConsumer(dbs, logger, settings)
-
-	if err := kafka.Consume(ctx, kc, cevConsumer.Process, logger); err != nil {
-		logger.Fatal().Err(err).Msg("Couldn't start event consumer.")
-	}
-
-	logger.Info().Msg("Contract events consumer started.")
+	//kc := kafka.Config{
```
elffjs, line 77: Let's not commit this commenting out!

graph/devicedefinition.resolvers.go
```diff
@@ -0,0 +1,21 @@
...
+// Code generated by github.com/99designs/gqlgen version v0.17.39
+
+import (
+	"context"
+
+	"github.com/DIMO-Network/identity-api/graph/model"
+)
+
+// Devicedefinition is the resolver for the devicedefinition field.
+func (r *queryResolver) Devicedefinition(ctx context.Context, by model.DevicedefinitionBy) (*model.DeviceDefinition, error) {
```
elffjs, line 14: We seem split in the Go and the GraphQL on whether it's `devicedefinition` or `deviceDefinition`. I would vote for `deviceDefinition`.

graph/resolver.go
```diff
@@ -71,10 +82,21 @@ type Resolver struct {
 	synthetic        SyntheticRepository
 	vehicle          VehicleRepository
 	vehicleprivilege vehicleprivilege.Repository
+	deviceDefinition DeviceDefinitionRepository
 }
 
 // NewResolver creates a new Resolver with allocated repositories.
 func NewResolver(baseRepo *base.Repository) *Resolver {
+
```
elffjs, line 90:
```suggestion
```

graph/schema/devicedefinition.graphqls
```diff
@@ -0,0 +1,166 @@
...
+  """
+  Device Type for this device definition.
+  """
+  deviceType: String
+
+  """
+  Image URI for this device definition.
+  """
+  imageURI: String
```
elffjs, line 75: Are some of these required?

graph/schema/devicedefinition.graphqls
```diff
@@ -0,0 +1,166 @@
...
+"""
+input DevicedefinitionFilter {
+
+  """
+  The manufacturer for the device definition.
+  """
+  manufacturer: String!
+
+  """
+  ID filters for the device definition that are of the given model.
```
elffjs, line 153: This description doesn't seem right, this seems like the description for model.

graph/schema/devicedefinition.graphqls
```diff
@@ -0,0 +1,166 @@
...
+  """
+  id: String
+
+  """
+  Model filters for device definition that are of the given model.
```
elffjs, line 158:
```suggestion
  Model filters for device definitions that are of the given model.
```

internal/helpers/helpers.go
```diff
@@ -44,3 +53,17 @@ func GetAftermarketDeviceImageUrl(baseURL string, tokenID int) string {
 func GetVehicleDataURI(baseURL string, tokenID int) string {
 	return fmt.Sprintf("%s%d", baseURL, tokenID)
 }
+
+func SlugString(term string) string {
+
```
elffjs, line 58:
```suggestion
```

internal/helpers/helpers.go
```diff
@@ -44,3 +53,17 @@ func GetAftermarketDeviceImageUrl(baseURL string, tokenID int) string {
 func GetVehicleDataURI(baseURL string, tokenID int) string {
 	return fmt.Sprintf("%s%d", baseURL, tokenID)
 }
+
+func SlugString(term string) string {
```
elffjs, line 57: What does this function do? It might be worth adding a comment and some tests. The meat seems to be line 62 but I can't tell offhand what it does.

internal/repositories/devicedefinition/devicedefinition_test.go
```diff
@@ -0,0 +1,39 @@
+package devicedefinition
+
+import (
+	"testing"
+)
+
+//const xmigrationsDir = "../../../migrations"
+
+func TestGetDeviceDefinition(t *testing.T) {
+	//ctx := context.Background()
```
elffjs, line 10: Do we mean to comment all this out?

graph/schema/devicedefinition.graphqls
```diff
@@ -0,0 +1,166 @@
...
+
+"""
+Filter for Device Definition.
+"""
+input DevicedefinitionFilter {
+
+  """
+  The manufacturer for the device definition.
+  """
+  manufacturer: String!
```
elffjs, line 150: I'm curious—why require manufacturer?

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	splitID := strings.Split(by.ID, "_")
+
+	if len(splitID) < 2 {
+		return nil, gqlerror.Errorf("The `ID` is incorrect.")
+	}
+
+	manufacturerSlug := splitID[0]
+
+	manufactures, err := r.ManufacturerCacheService.GetAllManufacturers(ctx)
```
elffjs, line 78:
```suggestion
	manufacturers, err := r.ManufacturerCacheService.GetAllManufacturers(ctx)
```

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,239 @@
...
+	if manufacturer == nil {
+		return nil, gqlerror.Errorf("Manufacturer %s doesn't exist.", manufacturerSlug)
+	}
+
+	tableName, err := r.ManufacturerContractService.GetTableName(ctx, manufacturer.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	statement := fmt.Sprintf("SELECT * FROM %s WHERE id = '%s'", *tableName, by.ID)
```
elffjs, line 100: Is this interpolation safe? I know it's a read-only query but it's still unnerving.

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,239 @@
...
+
+	if err = r.TablelandApiService.Query(ctx, queryParams, &modelTableland); err != nil {
+		return nil, err
+	}
+
+	for _, item := range modelTableland {
+		return ToAPI(manufacturerSlug, &item), nil
+	}
+
+	return nil, nil
```
elffjs, line 114: This confused me a bit—we expect this `modelTableland` to have length 1 or 0?

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,239 @@
...
+	if filterBy.ID != nil && len(*filterBy.ID) > 0 {
+		conditions = append(conditions, fmt.Sprintf("id = '%s'", *filterBy.ID))
+	}
+
+	whereClause := strings.Join(conditions, " AND ")
+	if whereClause != "" {
+		whereClause = " WHERE " + whereClause
+	}
+
+	manufactures, err := r.ManufacturerCacheService.GetAllManufacturers(ctx)
```
elffjs, line 143:
```suggestion
	manufacturers, err := r.ManufacturerCacheService.GetAllManufacturers(ctx)
```

graph/schema/devicedefinition.graphqls
elffjs: Does this have to be a "slug"? Can I write "Ford" here? Should we mention that in the doc?

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		return nil, gqlerror.Errorf("failed load manufactures: %s", err)
+	}
+
+	var manufacturer *services.ManufacturerCacheModel
```
elffjs, line 148: Perhaps the cache could offer retrieval by slug? We seem to do this in a couple of places.

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,239 @@
...
+	for _, model := range manufactures {
+		if model.Slug == filterBy.Manufacturer {
+			manufacturer = &model
+			break
+		}
+	}
+
+	tableName, err := r.ManufacturerContractService.GetTableName(ctx, manufacturer.ID)
```
elffjs, line 156: What if `manufacturer` is nil?

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	var modelCountTableland []DeviceDefinitionTablelandCountModel
+	if err = r.TablelandApiService.Query(ctx, countParams, &modelCountTableland); err != nil {
+		return nil, err
+	}
+
+	totalCount := modelCountTableland[0].Count
+
+	queryParams := map[string]string{
+		"statement": fmt.Sprintf("SELECT * FROM %s%s LIMIT %d OFFSET %d", *tableName, whereClause, limit, 1),
```
elffjs, line 172: Are we ordering this in any way? It doesn't feel like pagination is going to work. Why offset 1?

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,239 @@
...
+
+	countParams := map[string]string{
+		"statement": fmt.Sprintf("SELECT count(*) FROM %s%s", *tableName, whereClause),
+	}
+	var modelCountTableland []DeviceDefinitionTablelandCountModel
+	if err = r.TablelandApiService.Query(ctx, countParams, &modelCountTableland); err != nil {
+		return nil, err
+	}
+
+	totalCount := modelCountTableland[0].Count
```
elffjs, line 169: Is it safe?

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,239 @@
...
+			HasNextPage:     hasNext,
+			HasPreviousPage: hasPrevious,
+			StartCursor:     startCur,
+		},
+		TotalCount: int(totalCount),
+	}
+	if errList != nil {
+		return res, errList
+	}
+	return res, nil
```
elffjs, line 238: We'd get the same result by writing `return res, errList` yeah?

internal/services/contracts_manufacturer_service.go
```diff
@@ -0,0 +1,51 @@
...
+	return &ManufacturerContractService{
+		log:      log,
+		settings: settings,
+		client:   client,
+	}
+}
+
+func (m *ManufacturerContractService) GetTableName(ctx context.Context, manufacturerID int) (*string, error) {
+	contractAddress := common.HexToAddress(m.settings.DIMORegistryAddr)
+	queryInstance, err := contracts.NewRegistry(contractAddress, m.client)
```
elffjs, line 36: I think we could initialize these two things just once, in the `New` function.

internal/services/tableland_api_service.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+		fullURL.RawQuery = values.Encode()
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fullURL.String(), nil)
+	if err != nil {
+		return fmt.Errorf("failed to create request: %w", err)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
elffjs, line 47: Perhaps we should use a client with a timeout, or use a context with some deadline.

internal/services/tableland_api_service.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to complete request: %w", err)
+	}
+
+	if err != nil {
+		return err
+	}
```
elffjs, line 54: Why two of these?

internal/services/tableland_api_service.go
```diff
@@ -0,0 +1,62 @@
...
+
+func NewTablelandApiService(log *zerolog.Logger, settings *config.Settings) *TablelandApiService {
+	return &TablelandApiService{
+		log:      log,
+		settings: settings,
+	}
+}
+
+func (r *TablelandApiService) Query(ctx context.Context, queryParams map[string]string, result interface{}) error {
+	fullURL, err := url.Parse(r.settings.TablelandAPIGateway)
```
elffjs, line 27: Maybe do this in the "constructor"?

elffjs (changes requested):

graph/schema/devicedefinition.graphqls
```diff
@@ -0,0 +1,166 @@
...
+type DeviceDefinition implements Node {
+  """
+  An opaque global identifier for this device definition.
+  """
+  id: ID!
+
+  """
+  Legacy ID for this device definition.
+  """
+  ksuid: String
```
0xdev22, line 55: Can we name this something else, maybe `LegacyID`. KSUID doesn't describe what the variable does.

0xdev22 (changes requested):

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+
+	if manufacturer == nil {
+		return nil, gqlerror.Errorf("Manufacturer %s doesn't exist.", manufacturerSlug)
+	}
+
+	tableName, err := r.ManufacturerContractService.GetTableName(ctx, manufacturer.ID)
+	if err != nil {
+		return nil, err
```
0xdev22, line 97: Can we wrap this error `fmt.Errorf("...", err.Error())`

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	if last != nil {
+		slices.Reverse(all)
+	}
+
+	var errList gqlerror.List
+	var endCur, startCur *string
+	//if len(all) != 0 {
+	//	//ec := helpers.IDToCursor(all[len(all)-1].ID)
```
0xdev22, line 204: Please remove comment

internal/services/cache_manufacturer_service.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (m *ManufacturerCacheService) GetAllManufacturers(ctx context.Context) ([]ManufacturerCacheModel, error) {
+	if manufacturers, ok := m.cache.Get("manufacturers"); ok {
+		return manufacturers.([]ManufacturerCacheModel), nil
+	}
+
+	manufacturers, err := models.Manufacturers().All(ctx, m.PDB.DBS().Reader)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
```
0xdev22, line 50: I don't think `.All()` returns this error (I'm not sure). You should rather check if the returned `manufacturers` variable is empty

internal/services/cache_manufacturer_service_test.go
```diff
@@ -0,0 +1,77 @@
...
+		o.T().Fatal(err)
+	}
+}
+
+// Test Runner
+func TestManufacturerCacheServiceTestSuite(t *testing.T) {
+	suite.Run(t, new(ManufacturerCacheServiceTestSuite))
+}
+
+func (o *ManufacturerCacheServiceTestSuite) Test_Manufacturer_All_Success() {
```
0xdev22, line 58: Two tests that I see missing here 

- Since you want to return an error when manufacturers are not found, can you write a test for that scenario too.
- Can we verify the cache data is actually being returned early as implemented [here](https://github.com/DIMO-Network/identity-api/pull/88/files#diff-c76fc5f9f192673157bcdbfeda344b15ec590ab9351baa73bc8fe209c857a8d9R44)

0xdev22 (changes requested):

graph/schema/devicedefinition.graphqls
elffjs: I would think we would follow JavaScript casing and do `imageUri`. What do we do elsewhere in this repo?

internal/config/settings.go
rhvivancoeffio: The variables was added to values-prod.yaml and secret.yaml

internal/repositories/devicedefinition/devicedefinition.go
rhvivancoeffio: The references was updated

internal/repositories/devicedefinition/devicedefinition.go
rhvivancoeffio: The model was updated

.gitignore
rhvivancoeffio: Updated!!

cmd/identity-api/main.go
rhvivancoeffio: Updated!!

graph/devicedefinition.resolvers.go
rhvivancoeffio: The definition was updated

graph/schema/devicedefinition.graphqls
rhvivancoeffio: The property was updated

internal/services/contracts_manufacturer_service.go
rhvivancoeffio: The func was updated

internal/services/tableland_api_service.go
rhvivancoeffio: The definition was updated

internal/services/tableland_api_service.go
rhvivancoeffio: The definition was updated

internal/repositories/devicedefinition/devicedefinition.go
rhvivancoeffio: The func was updated

elffjs:
Will take a look tomorrow. Sorry about the delay.

.gitignore
elffjs: I'll do it.

graph/schema/devicedefinition.graphqls
```diff
@@ -0,0 +1,171 @@
...
+type DeviceDefinition implements Node {
+  """
+  An opaque global identifier for this device definition.
+  """
+  id: ID!
+
+  """
+  Device Definition ID for this device definition.
+  """
+  deviceDefinitionID: String
```
elffjs, line 55: I would do `deviceDefinitionId`. See "`tokenId`" elsewhere.

graph/schema/devicedefinition.graphqls
```diff
@@ -0,0 +1,171 @@
...
+
+  """
+  Legacy ID for this device definition.
+  """
+  legacyID: String
```
elffjs, line 60: Similarly, `legacyId`.

internal/repositories/devicedefinition/devicedefinition_test.go
elffjs: Still curious about this!

internal/services/tableland_api_service.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/DIMO-Network/identity-api/internal/config"
+	"github.com/DIMO-Network/shared"
+	"github.com/rs/zerolog"
+)
+
+type TablelandApiService struct {
```
elffjs, line 15: I would name this `TablelandAPIService` and so one. Go casing.

internal/services/tableland_api_service.go
```diff
@@ -0,0 +1,54 @@
...
+func (r *TablelandApiService) Query(_ context.Context, queryParams map[string]string, result interface{}) error {
+
+	var queryString string = "api/v1/query?"
+	if len(queryParams) > 0 {
+		queryParamsList := make([]string, 0, len(queryParams))
+		for key, value := range queryParams {
+			queryParamsList = append(queryParamsList, key+"="+value)
+		}
+		queryString += strings.Join(queryParamsList, "&")
+	}
```
elffjs, line 40: I would use [`url.Values`](https://pkg.go.dev/net/url#Values) for this.

internal/services/contracts_manufacturer_service.go
```diff
@@ -0,0 +1,61 @@
...
+
+func NewManufacturerContractService(log *zerolog.Logger,
+	settings *config.Settings,
+	client *ethclient.Client) (*ManufacturerContractService, error) {
+
+	contractAddress := common.HexToAddress(settings.DIMORegistryAddr)
+	registryInstance, err := contracts.NewRegistry(contractAddress, client)
+
+	if err != nil {
+		return nil, fmt.Errorf("failed instance NewRegistry: %s", err)
```
elffjs, line 30:
```suggestion
		return nil, fmt.Errorf("failed instance NewRegistry: %w", err)
```

internal/services/contracts_manufacturer_service.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	return &ManufacturerContractService{
+		log:              log,
+		settings:         settings,
+		registryInstance: registryInstance,
+	}, nil
+}
+
+func (m *ManufacturerContractService) GetTableName(ctx context.Context, manufacturerName string) (*string, error) {
```
elffjs, line 40: In general, I think we try to not call the contract in this repo. Is there some event or something we can listen to instead of doing this?

graph/schema/devicedefinition.graphqls
```diff
@@ -0,0 +1,171 @@
...
+
+type DeviceDefinitionAttribute {
+  """
+  Name for this device definition.
+  """
+  name: String
+
+  """
+  Value for this device definition.
+  """
```
elffjs, line 96: Are these really optional?

graph/schema/devicedefinition.graphqls
```diff
@@ -0,0 +1,171 @@
...
+
+  """
+  Device Type for this device definition.
+  """
+  deviceType: String
+
+  """
+  Image URI for this device definition.
+  """
+  imageUri: String
```
elffjs, line 80: Are these all optional?

graph/resolver.go
elffjs: I will run around deleting these newlines.

internal/helpers/helpers.go
elffjs: Maybe some tests would be good too.

internal/repositories/devicedefinition/devicedefinition.go
elffjs: I will fix these for you.

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	splitID := strings.Split(by.ID, "_")
+
+	if len(splitID) < 2 {
+		return nil, gqlerror.Errorf("The `ID` is incorrect.")
+	}
+
+	manufacturerSlug := splitID[0]
+
+	manufacturers, err := r.ManufacturerCacheService.GetAllManufacturers(ctx)
```
elffjs, line 84: Why do we need to pull and cache these? We have a manufacturers table—can we store the information we need there, using chain events?

elffjs: Okay, one call you're making is to `getManufacturerIdByName`, but we already have this information in the `manufacturers` table. Let me think about the other part.

internal/services/contracts_manufacturer_service.go
elffjs: What I would do is listen for these events:

https://github.com/DIMO-Network/dimo-identity/blob/e16ed5a88ff61a975e7927db83f5f465eb3b1365/contracts/implementations/tableland/DeviceDefinitionTable.sol#L28-L36

We could then maintain a column in the manufacturers table that knows the NFT id for the Tableland table corresponding to a manufacturer. Then the table name is just `_{chainId}_{tokenId}`.

elffjs: @LorranSutter It would be nice if creating the table also emitted `ManufacturerTableSet`, don't you think? Then we'd only have to listen for one event.

graph/schema/devicedefinition.graphqls
elffjs: Still curious about this. I think this will prove difficult to use.

If we're requiring the manufacturer, maybe we make this a sub-query of the `Manufacturer` object?

elffjs (changes requested):

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,256 @@
...
+	queryParams := map[string]string{
+		"statement": statement,
+	}
+	var modelTableland []DeviceDefinitionTablelandModel
+
+	if err = r.TablelandApiService.Query(ctx, queryParams, &modelTableland); err != nil {
+		return nil, err
+	}
+
+	fmt.Print(statement)
```
elffjs, line 116: Let's turn this into a real log statement, or delete it, something.

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,256 @@
...
+
+	var cont = 1
```
elffjs, line 118: Aren't we using this to construct the "global id" on line 49, and it's always 1? That seems odd.

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,256 @@
...
+	for _, item := range modelTableland {
```
elffjs, line 119: We expect this to always have one element in it? This is a bit hard to follow.

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	var modelCountTableland []DeviceDefinitionTablelandCountModel
+	if err = r.TablelandApiService.Query(ctx, countParams, &modelCountTableland); err != nil {
+		return nil, err
+	}
+
+	totalCount := modelCountTableland[0].Count
+
+	queryParams := map[string]string{
+		"statement": fmt.Sprintf("SELECT * FROM %s%s", *tableName, whereClause),
```
elffjs, line 183: I feel like we should put some ordering on this if we're going to do pagination.

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,256 @@
...
+	var cont = 1
+	for _, item := range modelTableland {
+		item.Index = cont
+		return ToAPI(&item)
+	}
+
+	return nil, nil
+}
+
+func (r *Repository) GetDeviceDefinitions(ctx context.Context, first *int, after *string, last *int, before *string, filterBy *gmodel.DeviceDefinitionFilter) (*gmodel.DeviceDefinitionConnection, error) {
```
elffjs, line 127: We don't appear to be using `before` or `after` in a serious way. I feel like this needs more thought.

charts/identity-api/templates/secret.yaml
```diff
@@ -14,6 +14,9 @@ spec:
   - remoteRef:
       key: {{ .Release.Namespace }}/identity/db/pass
     secretKey: DB_PASSWORD
+  - remoteRef:
+      key: {{ .Release.Namespace }}/identity/ethereum/rpc_url
```
JamesReate, line 18: secrets have been created

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,256 @@
...
+	if manufacturer == nil {
+		return nil, gqlerror.Errorf("Manufacturer %s doesn't exist.", manufacturerSlug)
+	}
+
+	tableName, err := r.ManufacturerContractService.GetTableName(ctx, manufacturer.Name)
+	if err != nil {
+		return nil, fmt.Errorf("error: %w", err)
+	}
+
+	statement := fmt.Sprintf("SELECT * FROM %s WHERE id = '%s'", *tableName, strings.ToLower(by.ID))
```
elffjs, line 106: What if we used a query builder like this?

```
sql, _, err := goqu.From("_5_33").Where(goqu.Ex{"id": "audi"}).ToSQL()
```

https://go.dev/play/p/eQpeIKM31_s

internal/services/tableland_api_service.go
elffjs: Here's an example:

https://go.dev/play/p/p0H4lKPoQog

elffjs:
We will get the Tableland table id into the `manufacturers` table.

internal/services/contracts_manufacturer_service.go
```diff
@@ -0,0 +1,60 @@
...
+	manufacturerID, err := m.registryInstance.GetManufacturerIdByName(&bind.CallOpts{
+		Context: ctx,
+		Pending: true,
+	}, manufacturerName)
+
+	if err != nil {
+		return nil, gqlerror.Errorf("failed get GetManufacturerIdByName: %s", err)
+	}
+
+	tableName, err := m.registryInstance.GetDeviceDefinitionTableName(&bind.CallOpts{
```
elffjs, line 50: I think this function is unable to compute table names for tables that we didn't create. I would use [this API](https://docs.tableland.xyz/api/validator/get-table-information) to get the name. _That_ you can certainly cache.

elffjs: I added a `table_id` column to `manufacturers`, you can use that

https://github.com/DIMO-Network/identity-api/commit/021cb89e2e29a94bbdb8dfe59e00174393e5af75

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -0,0 +1,256 @@
...
+	if len(by.ID) == 0 {
+		return nil, gqlerror.Errorf("Provide exactly one `ID`.")
+	}
+	splitID := strings.Split(by.ID, "_")
+
+	if len(splitID) < 2 {
+		return nil, gqlerror.Errorf("The `ID` is incorrect.")
+	}
+
+	manufacturerSlug := splitID[0]
```
elffjs, line 82: This seems to suppose that all manufacturers will follow our id pattern. Is this something we're willing to do?

internal/services/cache_manufacturer_service.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil, err
+	}
+
+	var all []ManufacturerCacheModel
+	all = make([]ManufacturerCacheModel, 0)
+	for _, manufacturer := range manufacturers {
+		all = append(all, ManufacturerCacheModel{
+			ID:   manufacturer.ID,
+			Name: manufacturer.Name,
+			Slug: helpers.SlugString(manufacturer.Name),
```
elffjs, line 56: Do you want to just put these in the database?

internal/repositories/devicedefinition/devicedefinition.go
JamesReate: The ID format we're settling on is: `manufacturerSlug_modelSlug_year` with the underscore being the key separator for each section. Slugs for each part use `-` instead of spaces eg. `Land Rover` -> `land-rover`. Any contributor would need to follow our id pattern, yes. When we get to the point that external users start contributing, they can add metadata to help them identify / link DD's to their internal systems. Eg. edmunds uses a numeric id. 

@rhvivancoeffio let's add a comment on the expected format for this function (line 71).

charts/identity-api/templates/secret.yaml
elffjs: I don't think we'll need this.

graph/schema/devicedefinition.graphqls
elffjs: Also, if this is required, then I think we need to require `filterBy`. And in that case I think we should probably just put it underneath the manufacturer object?

internal/repositories/devicedefinition/devicedefinition.go
elffjs: Are we going to force them to use an empty prefix as well?

internal/repositories/devicedefinition/devicedefinition.go
elffjs: Are we going to force them to use an empty prefix as well?

graph/resolver.go
```diff
@@ -62,6 +64,14 @@ type VehicleRepository interface {
 	GetVehicle(ctx context.Context, id int) (*model.Vehicle, error)
 }
 
+// DeviceDefinitionRepository interface for mocking devicedefinition.Repository.
+//
+//go:generate mockgen -destination=./mock_devicedefinition_test.go -package=graph github.com/DIMO-Network/identity-api/graph DeviceDefinitionRepository
```
elffjs, line 69: Actually, why are we dumping all the mocks here? This seems undesirable.

internal/repositories/devicedefinition/devicedefinition.go
elffjs: We're hoping there won't be slug collisions?

internal/helpers/helpers.go
elffjs: It looks like there's a version of this in shared?

internal/repositories/devicedefinition/devicedefinition.go
JamesReate: Collisions would have to be for the same make model year, don't think that would be possible. If you're asking about collisions for the same model + year across all makes, that could maybe happen, but with all definitions in their own table per make not sure I follow how that would happen. 
Regarding the empty prefix for the table, I had been asking about this too. It would be nice to have a table prefix if this is possible, not sure if others care.

internal/helpers/pagination_helper.go
```diff
@@ -49,7 +49,7 @@ func ValidateFirstLast(first, last *int, maxPageSize int) (int, error) {
 			return 0, errors.New("the value for `first` cannot be negative")
 		}
 		if *first > maxPageSize {
-			return 0, fmt.Errorf("the value %d for `first` exceeds the limit %d", *last, maxPageSize)
+			return 0, fmt.Errorf("the value %d for `first` exceeds the limit %d", *first, maxPageSize)
```
elffjs, line 52: Whoops on my part.

elffjs:
It's sort of working now

image

I was doing edits on a branch and then accidentally pushed to yours and it's a long story. 

Here is a dump of the manufacturers on testnet:

[mfrs.csv](https://github.com/DIMO-Network/identity-api/files/15284922/mfrs.csv)

I do `\copy manufacturers from 'mfrs.csv' csv header` and this loads in.

elffjs:
Some things I notice, poking around

* Often `imageURI` is blank but not null. We should probably treat that like it's null.
* Attributes are often blank. We should probably filter those out.

internal/repositories/vehicle/vehicles.go
```diff
@@ -42,19 +43,27 @@ func (r *Repository) createVehiclesResponse(totalCount int64, vehicles models.Ve
...
-		imageURL, err := GetVehicleImageURL(r.Settings.BaseImageURL, dv.ID)
-		if err != nil {
-			wErr := fmt.Errorf("error getting vehicle image url: %w", err)
-			errList = append(errList, gqlerror.Wrap(wErr))
-			continue
+		var imageURI string
+
+		if dv.ImageURI.Valid {
+			imageURI = dv.ImageURI.String
+		} else {
```
Allyson-English, line 50: could we simplify this as:

```suggestion
		imageURI := dv.ImageURI.String
		if !dv.ImageURI.Valid {
```

I think this is done elsewhere as well

internal/repositories/vehicle/vehicles.go
```diff
@@ -249,20 +271,21 @@ func ToAPI(v *models.Vehicle, imageURL string, dataURI string) (*gmodel.Vehicle,
...
+			ID:    v.DeviceDefinitionID.Ptr(),
 			Make:  v.Make.Ptr(),
 			Model: v.Model.Ptr(),
 			Year:  v.Year.Ptr(),
 		},
 		ManufacturerID: v.ManufacturerID,
 		Name:           name,
-		Image:          imageURL,
+		ImageURI:       imageURI,
+		Image:          imageURI,
```
Allyson-English, line 282: should these both be the same?

internal/repositories/vehicle/vehicles.go
elffjs: I'm not sure how this image URI stuff got into the diff here. Unrelated.

elffjs: To be honest I just don't like touching `.String` if it's not valid. I know it's more lines. I'll make the change on main.

Allyson-English: yeah thats fair, just was thinking from a readability standpoint. I have an aversion to `else`'s

internal/repositories/vehicle/vehicles.go
elffjs: Yes. We want to get rid of `image` but mobile are using it.

internal/repositories/vehicle/vehicles.go
elffjs: Shouldn't appear in the diff anymore.

internal/repositories/vehicle/vehicles.go
Allyson-English: ah, got it... maybe we can add a comment that says `// in the process of deprecating this, mobile is still using it` or something similar

elffjs: See https://github.com/DIMO-Network/identity-api/blob/d53a4bff4ea0ede8c9fd2095db424c8a2ddcf920/graph/schema/vehicle.graphqls#L130

PR #90, opened 2024-04-24
SI-2582 Add manufacturerTokenId filter to vehicles query

elffjs:
Add an optional `manufacturerTokenId` filter to the `vehicles` query.

### Try it

1. Dump prod `pg_dump -h localhost -U <user> -T 'migrations*' -a identity_api > identity.sql`[^1]
2. Load it into `psql -h localhost -U dimo identity_api < identity.sql`
3. Run a query like
    ```graphql
    {
      vehicles(first: 10, filterBy: {manufacturerTokenId: 10}) {
        totalCount
      }
    }
    ```
    to see how many Audis there are.

### Future work

* I think manufacturer id should be a required field in the database and the GraphQL schema—[you can't create a vehicle without one](https://github.com/DIMO-Network/identity-api/blob/4aac3c1fedec102555242b93195aeba37fca5b05/internal/services/contracts_events_consumer.go#L198). That was an oversight. Instituting that will break a lot of stuff, particularly tests; I will try to get that done in a separate PR.
* It might be nice to have a `vehicles` query underneath the manufacturer query. That will take more work.

[^1]: Here `-T 'migrations*'` excludes the migrations table and sequence, and `-a` skips the create statements so that you only get the inserts.


graph/schema/vehicle.graphqls
```diff
@@ -72,6 +72,11 @@ input VehiclesFilter {
   Year filters for vehicles that are of the given year.
   """
   year: Int
+
+  """
+  Filter for vehicles produced by a particular manufacturer, specified by manufacturer token id.
+  """
+  manufacturerId: Int
```
elffjs, line 79: Now that I think about it, this should be `manufacturerTokenId`.

elffjs:
Should I close this xff

zer0stars:
Merge if ready!

Allyson-English (approved):

PR #92, opened 2024-04-30
Add handling for null earinings

KevinJoiner:
This PR fixes 3 related issues.

1. Our schema has fields with type `BigInt!` which maps to the go type *BigInt if we return a nil value GQL returns the error `the requested element is null which the schema does not allow`.

2. We had checks for TotalSum being 0, which prevented us from returning other relevant information, such as totalCount.

3. Our SUM for calculating TotalTokens was adding cols that could be `NULL`. Unlike the SUM function if you call operators like `+` on `NuLL` they will always return `NULL`.


elffjs:
I'm wondering whether we should be allowing null token amounts at all in that table. I feel like it doesn't add anything.

internal/repositories/reward/rewards.go
```diff
@@ -305,3 +281,11 @@ func (r *Repository) PaginateGetEarningsByUsersDevices(ctx context.Context, user
 	userDeviceEarnings.History.PageInfo = afd.PageInfo
 	return userDeviceEarnings.History, nil
 }
+
+// NullDecimalToInt converts a null decimal to a big int.
```
elffjs, line 285: Maybe mention that null becomes zero?

graph/rewards_test.go
```diff
@@ -1334,3 +1334,111 @@ func (r *RewardsQueryTestSuite) Test_Query_GetUserRewards_BackPaginate_LastBefor
...
+	// Aftermarket Earnings
+	adEarn, ok := new(big.Int).SetString("1", 10)
+	r.True(ok)
+
+	// Streak Earnings
+	strkEarn, ok := new(big.Int).SetString("1", 10)
+	r.True(ok)
+
+	var rewards = []models.Reward{
+		{
```
elffjs, line 1354: Might we explicitly write out the zero values here that would have tripped up the old code? It's a bit difficult to notice the omission, there are lots of fields.

graph/rewards_test.go
```diff
@@ -1334,3 +1334,111 @@ func (r *RewardsQueryTestSuite) Test_Query_GetUserRewards_BackPaginate_LastBefor
...
+
+func (r *RewardsQueryTestSuite) Test_Query_GetUserRewards_NullEarnings() {
+	currTime := time.Now().UTC().Truncate(time.Second)
+	_, beneficiary, err := test.GenerateWallet()
+	r.NoError(err)
+
+	r.createDependencies()
+
+	// Aftermarket Earnings
+	adEarn, ok := new(big.Int).SetString("1", 10)
```
elffjs, line 1346: Maybe `big.NewInt(1)` for something this small?

elffjs (approved):

elffjs (approved):

migrations/20240501124220_not_null_token_values.sql
```diff
@@ -0,0 +1,25 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
+
+ALTER TABLE rewards
+ALTER COLUMN streak_earnings SET DEFAULT 0,
```
elffjs, line 6: Ah crap I think you need to still do an `UPDATE rewards SET ...`. Trying to patch this up.

migrations/20240501124220_not_null_token_values.sql
```diff
@@ -0,0 +1,25 @@
+-- +goose Up
```
elffjs, line 1: Oh one more thing—we try to name this sequentially. I think goose has a command that helps you do that.

PR #95, opened 2024-05-06
Update makefile

KevinJoiner:


elffjs:
This looks like a good idea, thanks.

My knowledge of make is all copy paste so I'll just assume the syntax is great.

Allyson-English (approved):

elffjs:
By the way, Goose has a command `goose fix` that makes the migrations sequential.

elffjs:
Okay yeah check this out:

https://github.com/DIMO-Network/identity-api/blob/7341b00f13e9230342d14eb420d7c8af39524e6d/Makefile#L13

I think it got removed when I freaked out and replaced everything with the standard Benthos Makefile copy-paste because I couldn't get the container to build. But I think it was a good idea.

KevinJoiner:
ooh goose fix is cool. I added the `-s` to the make file, which will knew migration files sequential.

KevinJoiner:
@elffjs That help command is super cool!

elffjs (approved):

elffjs:
> ooh goose fix is cool. I added the `-s` to the make file, which will knew migration files sequential.

Didn't know that flag. That's even better.

PR #97, opened 2024-05-07
convert wei values to token

KevinJoiner:
Updated the code to return a decimal.Big with 2 decimal places.


internal/repositories/reward/rewards.go
```diff
@@ -8,13 +8,16 @@ import (
...
 	"github.com/DIMO-Network/identity-api/models"
+	"github.com/ericlagergren/decimal"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/volatiletech/null/v8"
 	"github.com/volatiletech/sqlboiler/v4/queries/qm"
 	"github.com/volatiletech/sqlboiler/v4/types"
 	"golang.org/x/exp/slices"
 )
 
+var weiVal = decimal.WithContext(decimal.Context{Precision: 20}).SetMantScale(1, -18)
```
elffjs, line 19: Maybe a name like `weiPerEther` and a comment about how it's `10^8` (I always get weirded out by the negative).

elffjs: Could you say why the precision matters? I would've naively done `decimal.New(1, -18)`, I think.

internal/repositories/reward/rewards.go
```diff
@@ -276,3 +283,8 @@ func (r *Repository) PaginateGetEarningsByUsersDevices(ctx context.Context, user
 	userDeviceEarnings.History.PageInfo = afd.PageInfo
 	return userDeviceEarnings.History, nil
 }
+
+// divide by 10^18 to get token value.
+func weiToToken(wei types.Decimal) *decimal.Big {
+	return new(decimal.Big).Quo(wei.Big, weiVal)
```
elffjs, line 289: My guess is the precision in `weiVal` is intended to give this max precision (the library appears to default to 16?), but in testing what appears to matter is the precision of the instance on which you're calling `Quo`.

https://go.dev/play/p/PjWoBg-HCWa

This is an extremely minor point. Feel free to ignore.

internal/repositories/reward/rewards_test.go
```diff
@@ -377,11 +378,11 @@ func (r *RewardsRepoTestSuite) Test_PaginateVehicleEarningsByID_FwdPagination_Fi
 				Week:                    2,
 				Beneficiary:             common.BytesToAddress(beneficiary.Bytes()),
 				ConnectionStreak:        &connStrk,
-				StreakTokens:            strkEarn,
+				StreakTokens:            bigWeiToToken(strkEarn),
```
elffjs, line 381: I wonder if we need to use `big.Int` at all in this repo anymore.

graph/schema/schema.graphqls
```diff
@@ -14,6 +14,11 @@ An integer of arbitrary precision, decimal-encoded. Typically a uint256.
 """
 scalar BigInt
```
elffjs, line 15: We could maybe kill off `BigInt` for now. Is anything else using it?

elffjs (approved):
Minor comments.

internal/repositories/reward/rewards.go
KevinJoiner: Yeah, you are precisely right I was trying to give a little more precision (IDKY) by setting it on weiVal. Though I think we can use the default of 16. Since we only need a precision of 11 `max token count == billion tokens (9 places) + 2 decimals`

internal/repositories/reward/rewards_test.go
elffjs: Oh, that's a legitimate one. Those are huge.

internal/repositories/reward/rewards.go
KevinJoiner: I don't think it really does since I am rounding the result.

graph/schema/schema.graphqls
```diff
@@ -14,6 +14,11 @@ An integer of arbitrary precision, decimal-encoded. Typically a uint256.
...
 
+"""
+A floating-point number
```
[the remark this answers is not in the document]
KevinJoiner: I just went off the package comment
> decimal implements arbitrary precision, decimal floating-point numbers, per the [General Decimal Arithmetic](http://speleotrove.com/decimal/) specification.

https://pkg.go.dev/github.com/ericlagergren/decimal#section-readme

internal/repositories/reward/rewards.go
elffjs: It's unclear to me that setting it to 20 on the denominator is _doing_ anything.

https://go.dev/play/p/06Xgl71EUGD

graph/schema/schema.graphqls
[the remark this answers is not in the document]
elffjs: I was hoping for the word decimal but I think I'm merely being annoying.

internal/repositories/reward/rewards.go
KevinJoiner: it's not and if it was I think the default would still suffice for our needs.

elffjs (approved):

PR #98, opened 2024-05-08
Require manufacturers for vehicles and aftermarkets

elffjs:
Break and fix essentially all tests involving vehicles and aftermarket devices.


elffjs:
Did a little safety check:

```
SELECT * FROM vehicles WHERE manufacturer_id IS NULL;
SELECT * FROM aftermarket_devices WHERE manufacturer_id IS NULL;
```

Both turn up empty on prod.

KevinJoiner (approved):
So many tests 😵‍💫

PR #99, opened 2024-05-09
Feature/si 2658 image uri

zer0stars:
- Index ImageURI attribute from the chain ```https://polygonscan.com/tx/0xc10fce2b6c7ae9367d07d0fc0f9809bf97db6215d3513cad3ad7e874c50af409#eventlog```
- Update image to imageURI in response
- Add imageUri column to vehicle tables


zer0stars:
@LorranSutter did we make a mistake here by doing ImageUri instead of ImageURI in the contract attributes? It hasn't been used so I think we can change it if needed.

elffjs:
We’re in TypeScript/Solidity land. I think `ImageUri` fits okay.

LorranSutter:
Well... it is kinda out of the pattern we've been following. We have `DataURI`, `DefinitionURI`, but `ImageUri`. It hurts me a bit. We can change.

elffjs:
That I didn't see. Yeah, we'd better change then.

graph/schema/vehicle.graphqls
```diff
@@ -126,7 +126,7 @@ type Vehicle implements Node {
   """
   The Image Url of he vehicle
   """
-  image: String!
+  imageUri: String!
```
elffjs, line 129: Didn't catch that the first time.

internal/repositories/aftermarket/aftermarket_devices.go
```diff
@@ -191,7 +191,7 @@ type aftermarketDevicePrimaryKey struct {
 	TokenID int
 }
 
-// ToAPI converts a vehicle to a corresponding API vehicle.
+// ToAPI converts a aftermarket device to a corresponding API aftermarket device.
```
elffjs: I'm going to fix the grammar here.

elffjs:
Rewrote some of this in #100. If you're willing to go further, in #101 we get rid of this URI construction function returning errors and things clean up a little.

elffjs (commented):
This will be poggers when it compiles.

internal/repositories/vehicle/vehicles.go
```diff
@@ -42,11 +42,16 @@ func (r *Repository) createVehiclesResponse(totalCount int64, vehicles models.Ve
 	nodes := make([]*gmodel.Vehicle, len(vehicles))
 
 	for i, dv := range vehicles {
-		imageURL, err := GetVehicleImageURL(r.Settings.BaseImageURL, dv.ID)
```
elffjs, line 45: We need to do this for the singular get too.

elffjs:
Test failure appears to be some Testcontainers thing. I think this does what you want now.

KevinJoiner (approved):
tests passed after a re-run

elffjs (approved):

PR #107, opened 2024-05-24
remove dynamic properties, stuff that isn't hashed out

JamesReate:
remove:
- attributes
- imageUri


JamesReate:
seeing some unrelated tests fail in CI, but not locally. gqlgen was run to regen the models.

elffjs:
It's fine. I did some filtering.

PR #108, opened 2024-06-10
SI-2701-filter-vehicles-by-dd 

Allyson-English:


zer0stars:
Should this be by ksuid or readable string?

elffjs:
It’ll end up being the semi-readable new strings. No KSUIDs in this API.

Allyson-English:
updated the tests

elffjs:
I would use real DD ids for your dummy strings. You can query for these from identity!

elffjs (approved):

PR #109, opened 2024-06-11
SI-2707-vehicle-by-sd

Allyson-English:
find vehicle a synthetic device is connected to

### DIFF
--- a/graph/schema/synthetic.graphqls
+++ b/graph/schema/synthetic.graphqls
@@ -65,7 +65,7 @@ type SyntheticDevice implements Node {
   """
   mintedAt: Time!
   """
-  The vehicle, if any, with which the synthetic device is paired.
+  The vehicle with which the synthetic device is paired.
   """
   vehicle: Vehicle!
 }
--- a/graph/schema/synthetic.graphqls
+++ b/graph/schema/synthetic.graphqls
@@ -64,6 +64,10 @@ type SyntheticDevice implements Node {
   The block timestamp at which this device was minted.
   """
   mintedAt: Time!
+  """
+  The vehicle, if any, with which the synthetic device is paired.
+  """
+  vehicle: Vehicle
 }
 
 """
--- a/graph/schema/synthetic.graphqls
+++ b/graph/schema/synthetic.graphqls
@@ -67,7 +67,7 @@ type SyntheticDevice implements Node {
   """
   The vehicle, if any, with which the synthetic device is paired.
   """
-  vehicle: Vehicle
+  vehicle: Vehicle!
 }
 
 """
--- a/internal/repositories/synthetic/synthetic_devices.go
+++ b/internal/repositories/synthetic/synthetic_devices.go
@@ -92,7 +92,9 @@ func (r *Repository) GetSyntheticDevices(ctx context.Context, first *int, last *
 			return nil, fmt.Errorf("invalid after cursor: %w", err)
 		}
 		queryMods = append(queryMods, models.SyntheticDeviceWhere.ID.LT(afterID))
-	} else if before != nil {
+	}
+
+	if before != nil {
 		beforeID, err := helpers.CursorToID(*before)
 		if err != nil {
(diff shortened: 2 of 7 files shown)


KevinJoiner (approved):

graph/schema/synthetic.graphqls
```diff
@@ -64,6 +64,10 @@ type SyntheticDevice implements Node {
   The block timestamp at which this device was minted.
   """
   mintedAt: Time!
+  """
+  The vehicle, if any, with which the synthetic device is paired.
+  """
+  vehicle: Vehicle
```
elffjs, line 70: Probably `Vehicle!` for now?

internal/repositories/synthetic/synthetic_devices.go
```diff
@@ -91,25 +92,23 @@ func (r *Repository) GetSyntheticDevices(ctx context.Context, first *int, last *
 			return nil, fmt.Errorf("invalid after cursor: %w", err)
 		}
 		queryMods = append(queryMods, models.SyntheticDeviceWhere.ID.LT(afterID))
-	}
-
-	if before != nil {
+	} else if before != nil {
```
elffjs, line 95: Both can be specified. That's legal.

internal/repositories/synthetic/synthetic_devices.go
```diff
@@ -91,25 +92,23 @@ func (r *Repository) GetSyntheticDevices(ctx context.Context, first *int, last *
...
 		beforeID, err := helpers.CursorToID(*before)
 		if err != nil {
 			return nil, fmt.Errorf("invalid before cursor: %w", err)
 		}
 		queryMods = append(queryMods, models.SyntheticDeviceWhere.ID.GT(beforeID))
 	}
 
-	orderBy := "DESC"
+	orderBy := " DESC"
```
elffjs, line 103: Is this a good change?

elffjs:
@Allyson-English Giving this a nudge.

Allyson-English:
@elffjs thanks for nudging! I made those edits on friday but forgot to push them up because I got distracted by cases of if/ else for the before/ after queries. I removed the use of it here because you pointed out that it's legal but noticed that we're also doing it in these spots and was wondering if we wanted to use this opportunity to make both options allowable in those instances too: [1](https://github.com/DIMO-Network/identity-api/blob/main/internal/repositories/aftermarket/aftermarket_devices.go#L82), [2](https://github.com/DIMO-Network/identity-api/blob/main/internal/repositories/dcn/dcn.go#L127), [3](https://github.com/DIMO-Network/identity-api/blob/main/internal/repositories/reward/rewards.go#L83)

elffjs:
That's a good catch. I can do that in another PR, no worries.

I don't think it's a very _useful_ option but unlike combining first and last it doesn't seem to do any harm.

internal/repositories/synthetic/synthetic_devices.go
```diff
@@ -43,6 +43,7 @@ func ToAPI(sd *models.SyntheticDevice) (*gmodel.SyntheticDevice, error) {
 		IntegrationID: sd.IntegrationID,
 		Address:       common.BytesToAddress(sd.DeviceAddress),
 		MintedAt:      sd.MintedAt,
+		VehicleID:     &sd.VehicleID,
```
elffjs, line 46: We need to regenerate, I think? This isn't a pointer anymore.

elffjs (changes requested):

graph/schema/synthetic.graphqls
```diff
@@ -64,6 +64,10 @@ type SyntheticDevice implements Node {
   The block timestamp at which this device was minted.
   """
   mintedAt: Time!
+  """
+  The vehicle, if any, with which the synthetic device is paired.
```
elffjs, line 68: I think we can remove the `if any` here.

internal/repositories/synthetic/synthetic_devices.go
elffjs: (I do not think it is)

PR #112, opened 2024-07-11
SI-2717-index-acct-signers

Allyson-English:
This adds an event listener to map a deployed ECDSA signer to a wallet address and allows a user to query accounts and see associated kernel addresses


settings.sample.yaml
```diff
@@ -18,4 +18,5 @@ SYNTHETIC_DEVICE_CONTRACT_ADDRESS: "0x85226A67FF1b3Ec6cb033162f7df5038a6C3bAB2"
 REWARDS_CONTRACT_ADDRESS: "0x375885164266d48C48abbbb439Be98864Ae62bBE"
 BASE_IMAGE_URL: "https://devices-api.dev.dimo.zone/v1/"
 TABLELAND_API_GATEWAY: "https://testnets.tableland.network/"
-ETHEREUM_RPC_URL: "http://127.0.0.1:8545"
\ No newline at end of file
+ETHEREUM_RPC_URL: "http://127.0.0.1:8545"
+KERNAL_ECDSA_VALIDATOR: "0x8104e3Ad430EA6d354d013A6789fDFc71E671c43"
```
Allyson-English, line 22:
```suggestion
KERNEL_ECDSA_VALIDATOR: "0x8104e3Ad430EA6d354d013A6789fDFc71E671c43"
```

internal/repositories/account/account.go
```diff
@@ -0,0 +1,161 @@
...
+
+type AccountCursor struct {
+	OwnerAddress []byte
+	Kernel       []byte
+}
+
+// ToAPI converts a Account database row to a Account API model.
+func ToAPI(d *models.KernelAccount) (*gmodel.Account, error) {
+	return &gmodel.Account{
+		Kernel: []common.Address{common.BytesToAddress(d.Kernel)},
```
elffjs, line 31: I don't understand why this is a list that's always length 1.

gqlgen.yml
```diff
@@ -181,3 +181,7 @@ models:
         type: "*int"
       VehicleID:
         type: "int"
+  Account:
+    extraFields:
+      kernel:
+        type: "github.com/ethereum/go-ethereum/common.Address"
```
elffjs, line 187: Do we do anything with this?

internal/repositories/account/account.go
```diff
@@ -0,0 +1,161 @@
...
+	*base.Repository
+}
+
+type AccountCursor struct {
+	OwnerAddress []byte
+	Kernel       []byte
+}
+
+// ToAPI converts a Account database row to a Account API model.
+func ToAPI(d *models.KernelAccount) (*gmodel.Account, error) {
```
elffjs, line 29: Also, this can't error so maybe take that out of the signature?

internal/repositories/account/account.go
```diff
@@ -0,0 +1,264 @@
...
+		return r.GetAccountAddress(ctx, *by.Address), nil
+	}
+
+	return nil, gqlerror.Errorf("Provide exactly one of `signer`, `kernel` or `address`.")
+}
+
+func (r *Repository) GetAccountBySigner(ctx context.Context, signer common.Address) *gmodel.Account {
+	all, err := models.KernelAccounts(models.KernelAccountWhere.OwnerAddress.EQ(signer.Bytes())).All(ctx, r.PDB.DBS().Reader)
+	if err != nil {
+		return nil
```
KevinJoiner, line 49: I think it would be good to log this if it is any error other than `sql.ErrNoRows`

internal/repositories/account/account.go
```diff
@@ -0,0 +1,264 @@
...
+				Address:     common.BytesToAddress(signer.Bytes()),
+				SignerAdded: acct.SignerAdded,
+			},
+		})
+	}
+
+	return apiAccount
+}
+
+func (r *Repository) GetAccountByKernel(ctx context.Context, kernel common.Address) *gmodel.Account {
```
KevinJoiner, line 70: nit: These three functions could combined be a single function that takes in a `qm.QueryMod`

internal/repositories/account/account.go
```diff
@@ -0,0 +1,264 @@
...
+				Address:     common.BytesToAddress(kernel.Bytes()),
+				SignerAdded: acct.SignerAdded,
+			},
+		})
+	}
+
+	return &apiAccount
+}
+
+func (r *Repository) GetAccountAddress(ctx context.Context, address common.Address) *gmodel.Account {
```
KevinJoiner, line 94: nit:
```suggestion
func (r *Repository) GetAccountByAddress(ctx context.Context, address common.Address) *gmodel.Account {
```

internal/repositories/account/account.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	queryMods := []qm.QueryMod{}
+	if filterBy != nil && filterBy.Signer != nil {
+		queryMods = append(queryMods, models.KernelAccountWhere.OwnerAddress.EQ(filterBy.Signer.Bytes()))
+	}
+
+	accountCount, err := models.KernelAccounts(queryMods...).Count(ctx, r.PDB.DBS().Reader)
+	if err != nil {
+		return nil, err
```
KevinJoiner, line 136: We can log this and return a generic internal error so we do not return DB errors to the user.

internal/repositories/account/account.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	orderBy := " DESC"
+	if last != nil {
+		orderBy = " ASC"
+	}
+
+	queryMods = append(queryMods,
+		qm.Limit(limit+1),
+		qm.OrderBy(models.KernelAccountColumns.OwnerAddress+orderBy+", "+models.KernelAccountColumns.Kernel+orderBy),
```
KevinJoiner, line 146: This is fine just showing you an alternative syntax
```suggestion
		qm.OrderBy(models.KernelAccountColumns.OwnerAddress+orderBy),
		qm.OrderBy(models.KernelAccountColumns.Kernel+orderBy),
```

KevinJoiner (changes requested):

PR #113, opened 2024-07-22
[SM-994] make model filter case insensitive

Asfiroth:
- Make tableland model filter case insensitive.

### DIFF
--- a/internal/repositories/devicedefinition/devicedefinition.go
+++ b/internal/repositories/devicedefinition/devicedefinition.go
@@ -138,7 +138,7 @@ func (r *Repository) GetDeviceDefinitions(ctx context.Context, tableID, first *i
 		}
 
 		if filterBy.Model != nil {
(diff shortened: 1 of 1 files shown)


Allyson-English (approved):

JamesReate (approved):
let's try it out

elffjs:
Have you tested this? Can you present any evidence that it works?

elffjs:
If it does indeed work: We should update the docs on the filter to tell people.

Asfiroth:
Yup, this query: `https://testnets.tableland.network/api/v1/query?statement=select * from _80002_8 where model = '4c' COLLATE NOCASE` returns data
Screen Shot 2024-07-23 at 14 18 06 PM

Same query without `COLLATE NOCASE` returns nothing
Screen Shot 2024-07-23 at 14 19 19 PM

since the exacto Model for Alfa Romeo would be '4C'  with first query is working with '4c'

elffjs:
I want to see it work in the context of identity-api.

We merged it anyway so I guess I can test myself.

PR #118, opened 2024-08-20
AS-2923-sacd

Allyson-English:
add test

### DIFF
--- a/internal/repositories/vehiclesacd/vehicle_sacd_test.go
+++ b/internal/repositories/vehiclesacd/vehicle_sacd_test.go
@@ -1,0 +1,157 @@
+package vehiclesacd
+
+import (
+	"context"
+	"fmt"
+	"math/big"
+	"testing"
+	"time"
+
+	"github.com/DIMO-Network/identity-api/graph/model"
+	"github.com/DIMO-Network/identity-api/internal/config"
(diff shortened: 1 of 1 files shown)


internal/repositories/vehiclesacd/vehicle_sacd_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+	"github.com/testcontainers/testcontainers-go/modules/postgres"
+	"github.com/volatiletech/null/v8"
+	"github.com/volatiletech/sqlboiler/v4/boil"
+)
+
+type VehiclesSacdRepoTestSuite struct {
+	suite.Suite
+	ctx       context.Context
```
KevinJoiner, line 27: One weird rule is that you should not have context as member variables. This is a test so I really don't care about it just thought you might be interested to know.

>Don’t add a Context member to a struct type; instead add a ctx parameter to each method on that type that needs to pass it along. The one exception is for methods whose signature must match an interface in the standard library or in a third party library.

https://go.dev/wiki/CodeReviewComments#contexts

KevinJoiner (approved):

PR #122, opened 2024-10-01
Sm 1189 add manufacturer to device definitions identity api

JamesReate:
- Add Manufacturer to device definition by id


JamesReate:
test failures are due to postgres containers not spinning up or down on time

elffjs:
I just run it until it works.

README.md
```diff
@@ -2,6 +2,12 @@
 
 Run `make help` to see some helpful sub-commands:
 
+## If you're using MacOS, install latest make
```
elffjs, line 5: No need for this following #121

internal/repositories/devicedefinition/devicedefinition.go
```diff
@@ -254,7 +262,7 @@ func (r *Repository) GetDeviceDefinitions(ctx context.Context, tableID, first *i
 	nodes := make([]*gmodel.DeviceDefinition, len(all))
 
 	for i, dv := range all {
-		gv, err := ToAPI(&dv)
```
elffjs, line 257: So it won't show up in the list view? I know it's not terribly useful in that case because the manufacturer is the parent, but it seems inconsistent, I don't like it.

JamesReate: sounds good will add to the list too

PR #123, opened 2024-10-11
Add device lookup by IMEI and devEUI.

KevinJoiner:
A lot of files go touched because I ran generate with the correct gql version.


elffjs:
I normally do that in a separate commit and push to main.

elffjs (approved):

KevinJoiner:
I didn't update the gql version just the last person didn't run generate

elffjs:
What happens is I update go.mod and don't regenerate like a rat.

PR #139, opened 2025-02-24
Update connection settings

KevinJoiner:


KevinJoiner:
we used different values

PR #149, opened 2025-05-05
Add filter for manufacturer by slug

JamesReate:
gqlgqn version v0.17.73


KevinJoiner (approved):
We should update gqlgen in go.mod too if we are updating the generated code.

JamesReate:
> We should update gqlgen in go.mod too if we are updating the generated code.

done

PR #161, opened 2025-07-18
EX-3767 Add vehicle storage nodes

elffjs:
Handle events from the [storage node contract](https://github.com/DIMO-Network/storage-node/blob/a7ad186f121c7123be3d6f9899669dacd0537d92/src/StorageNode.sol#L30):

* `StorageNodeAnchorMinted`
* `Transfer`
* `NodeUriUpdated`
* `NodeSetForVehicle`

and display the sum of these in the API.

A few things to note:

* There are many similarities to the connections work:
  * The token ids are humongous, because they are formed by taking the human readable string `label` and computing `uint256(keccak256(bytes(label)))`.
  * A `bytes32` is the most convenient primary key, and we expend a lot of effort translating to and from this type.
* Existing vehicles don't have any node set, so the `storageNode` field on `Vehicle` may be null.
* There's already a [GraphQL `Node`](https://graphql.org/learn/global-object-identification/) concept, so it's easy to get confused in the API. We tried to always write the word “storage”.


graph/schema/storagenode.graphqls
```diff
@@ -0,0 +1,32 @@
...
+  The owner of the node. Nodes are transferable, so this may change over time.
+  """
+  owner: Address!
+  """
+  The token id of the node as an NFT. Since this is uint256(keccak256(bytes(label)))
+  it tends to be very large.
+  """
+  tokenId: BigInt!
+  """
+  The URI for the node. This will host the well-known URIs that tell clients how to send in
```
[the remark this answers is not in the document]
zakharenkodmytro: I see, so it is derived from the contract.

internal/services/node/node.go
```diff
@@ -0,0 +1,135 @@
+package node
```
zakharenkodmytro, line 1: same here, lmk if I misunderstand if we want be more strict about naming "node" vs "storage node" to avoid confusion

charts/identity-api/values.yaml
```diff
@@ -50,6 +50,7 @@ env:
   DEV_LICENSE_ADDR: '0xdb6c0dBbaf48b9D9Bcf5ca3C45cFF3811D70eD96'
   STAKING_ADDR: '0xA010a42DE263592E631d3DCDC621ae7BFfCca338'
   CONNECTION_ADDR: '0x41799E9Dc893722844E771a1C1cAf3BBc2876132'
```
zakharenkodmytro, line 52: once we merge, it will be live only on dev? For prod we will do separate PR later?

elffjs:
Tests failing here but not locally. Probably some race.

internal/repositories/node/node.go
```diff
@@ -0,0 +1,46 @@
+package node
```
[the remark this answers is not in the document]
elffjs: I prefer short stuff if it's at all possible. I debated this, I could change it, I haven't decided.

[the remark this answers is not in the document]
elffjs: I don't think dashed are allowed.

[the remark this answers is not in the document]
elffjs: I mostly meant _in the API_. Internally I'm just slapping this together. I'll probably do the rename, but it's just so damned verbose.

charts/identity-api/values.yaml
elffjs: My hope was that you guys could resolve any questions like, e.g., the `"uri"` naming, before we committed to something on prod.

zakharenkodmytro: ok, sounds reasonable to me

internal/services/node/node.go
elffjs: Okay I think we're good on these now.

PR #164, opened 2025-07-28
Read Definition MMY from definitions

JamesReate:
This is currently incomplete, but wanted to validate the initial approach first before doing more work on it -  updating the tests, linter and changing all usages of .ToAPI

Background:
The definition ID is what drives what definition the vehicle is set to. Most clients will use the Make, Model, Year in the vehicle response from identity-api to show users their MMY. In the past, we had thought having the MMY on the vehicle nft made sense, but lately seems to add complexity/issues. 
The objective of this PR is to solve for the user experience issues we've seen reported recently, although their may be a bigger underlying issue of changes not getting picked up correctly.

### DIFF
--- a/graph/schema.resolvers_test.go
+++ b/graph/schema.resolvers_test.go
@@ -25,7 +25,7 @@ func TestQueryResolver_Node(t *testing.T) {
 	baseRepo := &base.Repository{}
 
 	vehicleRepo := vehicle.New(baseRepo)
-	testVehicle, err := vehicleRepo.ToAPI(&models.Vehicle{ID: 1}, "", "")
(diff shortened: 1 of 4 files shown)


elffjs:
I would consider telling GQLGen that `definition` needs a resolver, and then pretty much all of the new code would be inside of a loader. Does that make sense? I think it would cut down on the lift here.

JamesReate:
> I would consider telling GQLGen that `Vehicle.definition` always needs a resolver, and then pretty much all of the new code (collecting definition ids, making the calls) would be inside of a loader. Does that make sense? I think it would cut down on the lift here.

I'll read up on GQLGen see how that works.

